PR #5, opened 2020-12-07
Add number format utility

Scoder12:
DRY up the code a bit by adding a utility function to format numbers in the reddit `{}k` format. 

Using an `i64` as input because scores can be negative, which requires convert the members `u64` which could be a potential overflow but is probably no problem.


spikecodes:
Thanks for the PR!

PR #6, opened 2020-12-07
Use rust-cache action and prettify workflow yml

Scoder12:
Use an action to cache rust packages to speed up compile times, and run `prettier` on the workflow file.


spikecodes:
Great PR!

PR #8, opened 2020-12-25
Add InsanityWtf Libreddit Instance

zachjmurphy:


spikecodes:
Thanks for the PR! The page gives me an error when I try to visit it:
image

spikecodes:
Error seems to be gone, thanks for the PR!

PR #19, opened 2021-01-01
Fix posts overflowing on Safari on iOS

somoso:
In Safari, the value `anywhere` is not supported for property `overflow-wrap`. Once changed to `break-word`, it behaves like it does in Chrome and Firefox.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -277,7 +277,7 @@ input[type="submit"]:hover { color: var(--accent); }
 .post_left, .post_right {
 	display: flex;
 	flex-direction: column;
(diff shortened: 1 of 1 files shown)


somoso:
I should probably add that I noticed this on my iPhone when looking at this post: https://libredd.it/r/hiphopheads/comments/kny16d/rip_mf_doom/ - on desktop it is fine (even in their "mobile mode" but when I investigated it in the iOS simulator on my Mac, I spotted the `anywhere` keyword wasn't supported).

spikecodes:
Thank you for the PR! Good catch, just merged!

PR #37, opened 2021-01-08
Fix theming for all browsers

somoso:
Really noticable on iOS, but ensuring all browsers get the love.

The buttons and input aren't as flat as they usually are on my desktop Firefox. This patch should sort that out.


spikecodes:
Awesome, thanks!

PR #38, opened 2021-01-08
Break word to stop it disappearing on mobile

somoso:
This kept happening to me but I couldn't reproduce it in the iPad Simulator. Finally got it nailed down and sorted.

Tested this on Safari (mobile and desktop), Firefox, and Edge browser.


spikecodes:
Thanks!

PR #46, opened 2021-01-12
Fix the thumbnail issue on iOS

somoso:
Thumbnails were ridiculously long.

Tested this change on Firefox, Safari, and Edge.


spikecodes:
Good fix, thanks!

PR #48, opened 2021-01-12
Relative timestamps for posts younger than 24h

robrobinbin:
Only affects the posts in the "subreddit" view, clicking the post still gives the full timestamp.

This is my first Rust project, so feedback is welcome.

image


spikecodes:
Thanks for this!

PR #57, opened 2021-01-16
Improve support for text-only browsers

robrobinbin:
I'm not sure whether this is a target you're interested in supporting, but this PR improves the rendering of libreddit on many text-only browsers (or browsers without CSS enabled in general).

Adds hr's between posts to separate them (hidden by CSS, so the view remains the same on a graphical browser).

Use blockquotes to indent nested comments (Again, no impact on a graphical browser as far as I'm aware).

The blockquote trick successfully works on `w3m`, `links2` and `elinks`. It didn't work on `lynx`.

# Screenshots (w3m)

## Subreddit view (before/after)

image image

## Comments (before/after)
image image

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -388,6 +388,10 @@ a.search_subreddit:hover {
 
 /* Post */
 
+.sep {
+	display: none;
+}
+
 .thread {
 	word-break: break-word;
 }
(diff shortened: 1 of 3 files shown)


spikecodes:
Great PR, thanks!

PR #62, opened 2021-01-18
Add "system" theme that matches browser/OS theme

mcrossman:
Adds a third theme option called `system` that matches the user's operating system theme if supported by the browser through the `prefers-color-scheme` media query. This is set as the new default. Choosing a specific theme will override this behaviour.

CSS currently uses two new sets of variables to define the themes and the base variables now choose which one to reference. The colour variables that used to be in general are now with everything else in the old light theme section.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -127,7 +127,7 @@ aside {
 	opacity: 0.5;
 }
 
-/* Color themes */
+/* Define themes */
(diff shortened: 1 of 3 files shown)


spikecodes:
Fantastic contribution, thanks!

PR #64, opened 2021-01-20
Should fix the strange overflowing placeholder when an image fails to load

robrobinbin:
...for some reason


spikecodes:
Thanks!

PR #68, opened 2021-01-21
Absolutly no jumping when images (fail to) load

robrobinbin:
Maybe I'm making all of this too complex 😅

* Tested on firefox and chrome
* Images have a max-height of 512px (expect when they are much longer than wide, because then they would become illegible)
* Images are clickable (direct link to image, for zooming)
* A placeholder image for loading/failed to load images (Chrome sadly overwrites it with an ugly, blurry icon though)

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -522,30 +522,21 @@ a.search_subreddit:hover {
 }
 
 .post_media {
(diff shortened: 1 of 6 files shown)


spikecodes:
Thanks!

PR #70, opened 2021-01-23
Experiment with caching requests to Reddit

robrobinbin:
Cache the JSON responses from Reddit, 

Current parameters:
* Max 1000 items in the cache
* Keep responses for max. 120 seconds
* Only `Ok` results are cached.

https://github.com/jaemk/cached


spikecodes:
Thanks! This is a good suggestion. Though, 120 seconds for caching seems like it might be a bit long, especially if one wants to see live updates. Perhaps it could be shorter? 30 or 60 seconds?

robrobinbin:
I've made it 60, but feel free to edit it further.
Let's see whether this has a noticeable impact :)

Ideally, we could have different values for content that changes more quickly in a noticeable manner, such as `/new` pages, but this implementation uses just a simple decorator.

spikecodes:
Thanks!

Issue #23, opened 2021-01-03
Feature Request | Subscribe to subreddits


gerroon:
Hi
It would be nice of one can add favorite subs and posts within the app. some kind of side bar access would be good I think.

thanks

spikecodes:
I like this idea, good suggestion!

spikecodes:
Not sure how it can be done without accounts but perhaps through the use of cookies (see #22).

gerroon:
Hi

The reason I asked about it because as a user I do not want to use an account or pollute my browser bookmarks. It would be nice if it is implemented from the user point of view. Just a suggestion.

Thanks

spikecodes:
Libreddit supports multireddits as of 565f4f2 so you can actually grab a "multireddit of your subscriptions" [here](https://reddit.com/subreddits) (bottom right) and visit that in Libreddit:

Eg. https://libredd.it/r/archlinux+Monero+linux+bitcoin+Unexpected+unixporn

gerroon:
Thanks for the update, just did a pull and all seems broken?

https://i.imgur.com/OR3x11b.png

spikecodes:
> just did a pull

Do you mean you ran `git pull` on your own instance?

I visited the example link I sent above and it rendered as:
![image](https://user-images.githubusercontent.com/19519553/103496315-93154080-4df2-11eb-843b-96b21bf0597c.png)

gerroon:
I am using Docker docker pull spikecodes/libreddit

The prev version was like the one you have in your post, the current docker pull has broken pages like I posted above.

spikecodes:
Just tried docker pulling and it shows fine for me but I will be pushing an update shortly so perhaps that will fix it on your end.

gerroon:
Still broken, no idea. Something must have changed  that it is broken since all I did was to pull the new version. 

I can try to debug it if you can tell me how to get some logs.

spikecodes:
Are you able to open a new issue for this? Can you include a screenshot, your browser and your operating system?

PR #78, opened 2021-01-30
Subscribing to subreddits (favorites)

mcrossman:
Implements basic subreddit subscriptions/favourites (#23), managed through each sub's sidebar.

I've never written in Rust before so the code is pretty rough.

### Changes in code/Implementation details

- Subscription list is a comma-separated list of subreddit names, saved under the `subscriptions` cookie.
- Adding a sub to the list requires clicking the "subscribe" button in the sidebar
  - Subscribe is a GET request to `/r/{{ subreddit }}/subscribe/`, which will add that sub to the cookie
- Removing a sub requires clicking "unsubscribe" in the sidebar or in settings
  - Unsubscribe is a GET request to `/r/{{ subreddit }}/unsubscribe/` which will remove that sub from the `subscriptions` cookie or delete the cookie if it's the last one
  - GET is used instead of POST because I ran into strange behaviour when trying to loop over the unsubscribe forms in settings. It seems looping with forms results in the first form tag being ignored.

### Visual changes

- A new subscriptions dropdown (uses `details` now but I might test hover in the future)
  - If you've subscribed to a subreddit, it will appear
  - It's a vertical list similar in appearance to the sort bar
  - The active subreddit is highlighted in the accent colour
- Margins in the `main` element have been reduced to allow for an element to fit in above. The navbar padding has become `body` padding.
- Other elements near the subscriptions dropdown have had their styles tweaked.

### Sub list considerations

~~The sub list currently appears as an extra horizontal bar (in the style of the sort bar) under the navbar.~~

The sub list was previously a horizontal bar. Now it's a drop-down button that should stay at the top left of the main content or be centred under the nav bar on mobile.

If we place it inside the navbar, it makes it way too tall, especially on mobile. Other sub lists like on RES or teddit are part of the navbar but those bars aren't fixed.

~~Another idea is to use a dropdown menu for a vertical list like new Reddit.~~

Reddit has a drop-down button in the navbar, but it was a bit messy trying to fit the button (& a smaller symbol variant) with everything else in the mobile navbar.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -480,9 +480,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.5.0"
+version = "3.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f07aa6688c702439a1be0307b6a94dffe1168569e45b9500c1372bc580740d59"
+checksum = "099e596ef14349721d9016f6b80dd3419ea1bf289ab9b44df8e4dfd3a005d5d9"
 
 [[package]]
 name = "byteorder"
@@ -531,9 +531,9 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chunked_transfer"
-version = "1.3.0"
+version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7477065d45a8fe57167bf3cf8bcd3729b54cfcb81cca49bda2d038ea89ae82ca"
+checksum = "fff857943da45f546682664a79488be82e69e43c1a7a2307679ab9afb3a66d2e"
 
 [[package]]
 name = "const_fn"
@@ -628,9 +628,9 @@ dependencies = [
 
 [[package]]
 name = "flate2"
-version = "1.0.19"
+version = "1.0.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7411863d55df97a419aa64cb4d2f167103ea9d767e2c54a1868b7ac3f6b47129"
+checksum = "cd3aec53de10fe96d7d8c565eb17f2c687bb5518a2ec453b5b1252964526abe0"
 dependencies = [
  "cfg-if 1.0.0",
  "crc32fast",
@@ -988,9 +988,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.83"
+version = "0.2.84"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7eb0c4e9c72ee9d69b767adebc5f4788462a3b45624acd919475c92597bcaf4f"
+checksum = "1cca32fa0182e8c0989459524dc356b8f2b5c10f1b9eb521b7d182c03cf8c5ff"
 
 [[package]]
 name = "libreddit"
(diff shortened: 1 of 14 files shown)


mcrossman:
Changed the sub list to be a dropdown more in line with new Reddit's vertical list.

libreddit-subs

robrobinbin:
Wow, very cool stuff and seems to work well!

The alignment looks a bit off on front page though ;)
image

robrobinbin:
The following is not a direct feedback on your PR or request to you, but rather just me brainstorming out loud about possible future iterations of this :smile:

I've had look at how the new reddit handles this currently and made the following mock up.

The idea would be to add a cog and info icon to settings and about, respectively, and hide the labels on mobile.

image

mcrossman:
> The alignment looks a bit off on front page though ;)

The way I've been aligning the button is a real mess.

> I've had look at how the new reddit handles this currently and made the following mock up.

On mobile, I'm thinking about hiding the logo and having three buttons above the search bar. Though it might be better to hide the search bar under a search button just like the sub list.

mcrossman:
Moved the sub list back into the navbar. 

I think an about page might be more appropriate in a future PR. This one's already getting quite large.

spikecodes:
Fantastic PR! Thanks! I'm able to subscribe successfully but I cannot unsubscribe for some reason. Are you able to reproduce this?

mcrossman:
> Fantastic PR! Thanks! I'm able to subscribe successfully but I cannot unsubscribe for some reason. Are you able to reproduce this issue?

Yes. This happens when you're only subscribed to one subreddit. I'm looking into it.

mcrossman:
```rust
if sub_list.is_empty() {
  res.del_cookie(&Cookie::named("subscriptions"));
  // cookie still exists after this
}
```

It seems that deleting a cookie in this way (which is based on `settings.rs`) doesn't seem to work. I feel like I'm missing a method.

The quick workaround is to get rid of the if statement and let it set an empty cookie.

spikecodes:
I'm not sure why the that method isn't working yet it works in settings. I also found that when unsubscribing from a subreddit, it sets the entire `subscriptions` cookie blank.
image

I will do some testing.

spikecodes:
An alternative for now would be setting the cookie, but simply making it's value blank. This isn't ideal (still takes up a cookie on the user's end) but would work.

mcrossman:
It works when I build a new cookie with the path included like this:

```rust
res.del_cookie(&Cookie::build("subscriptions", "").path("/").finish());
```

But when I exclude the path, it doesn't get deleted. I'm guessing that since we're in `/r/sub/unsubscribe` the default response cookie is not modifying cookies at the root path.

Returning a cookie with no value is how it gets deleted. That's what I see in the response cookies when the settings page tries to do it.

spikecodes:
Yeah, the way cookies get deleted is when you set the "expire" time to a past date or when the Max-Age is reached (`.del_cookie` sets the Max-Age to 0). You're right about the path. It seems that the `del_cookie` method seems to not specify one.

spikecodes:
Could 
```rust
if sub_list.is_empty() {
	sub_list = Vec::new();
	sub_list.push(sub_name);
} else if !sub_list.contains(&sub_name) {
	sub_list.push(sub_name);
	sub_list.sort();
}
```
become
```rust
sub_list.push(sub_name);
sub_list.sort();
}
```
? Because when a vector `is_empty()`, it's value is the same as `Vec::new()`.

mcrossman:
Yep, it can. Thanks.

spikecodes:
The default homepage is now a subscription feed! (or `r/popular` if there you aren't subscribed to any)
image

spikecodes:
It would nice to have a "Subscribe" button above the sort box when viewing multireddits (like `r/rust+privacy`) to allow people to subscribe to a multireddit of their Reddit subscriptions. This would make it easier for people to migrate from Reddit to Libreddit.

mcrossman:
Done my testing, too.

Looks good to merge.

spikecodes:
Thank you for this useful PR!

PR #82, opened 2021-01-31
Minor style tweak

robrobinbin:
on small screens, subscriptions are put below the logo, which looks a bit strange and makes it somehow harder to hit


spikecodes:
Thanks!

PR #83, opened 2021-02-01
iOS "Add to Homescreen" functionality

JPyke3:
# Introduction

iOS and other browsers support PWA's, this allows them to act similiar to native apps. on iOS, when you add the app to your homescreen, you are adding it as a PWA, and the App acts more like a native app than a website. This PR adds that functionality.

## Disclaimer

This does not add PWA support to Chrome, in order to get something like that to work you need a Service Worker, which only works if you add JavaScript, something this application clearly doesn't want to support. Furthermore, there would be privacy concerns with caching on device.

1) This pull simply allows the site to be added to iOS, and nothing is cached.
2) This functionality generally only works on HTTPS

## Video

https://user-images.githubusercontent.com/13283054/106445593-7dda1300-647f-11eb-86af-668ccd4ab698.MP4


spikecodes:
Awesome, thank you!

Mennaruuk:
You guys freakin' rock!

PR #92, opened 2021-02-02
Fix layout on browsers with faulty "display: contents" support (Safari)

robrobinbin:


spikecodes:
Fantastic PR, thanks so much!

PR #93, opened 2021-02-03
Make feed sorting case-insensitive

mcrossman:
Sorting is currently case-sensitive in that it puts subs starting with an upper-case letter ahead of ones that start with a lower-case letter.

To reproduce:

- Subscribe to r/LineageOS
- Subscribe to r/fossdroid

LineageOS is placed before fossdroid.


spikecodes:
Thanks!

PR #97, opened 2021-02-04
Minor update to README.md

Mennaruuk:
Changed letters to numbers for installation methods. That way, the hyperlinks at the top linking to the installation methods work. Super minor update!


spikecodes:
:smile: Awesome, thanks! I didn't catch those

Issue #112, opened 2021-02-09
Feature Request | Autofocus Comments


Mennaruuk:
**Is your feature request related to a problem? Please describe.**
Sometimes, I encounter direct links to a parent or child comment. The way Libreddit currently does it is it shows the comment in question in the vicinity of its parent/child comments. Unfortunately, it can be confusing.

For example, this is [the link ](https://libredd.it/r/StarWarsBattlefront/comments/7cff0b/seriously_i_paid_80_to_have_vader_locked/dpqe377/?utm_source=reddit&utm_medium=web2x&context=3)to this self-highlighted comment:
![image](https://user-images.githubusercontent.com/52135169/107407427-a5605980-6ad7-11eb-9320-7ebe08db8e4b.png)

[Another link](https://libredd.it/r/StarWarsBattlefront/comments/7cff0b/seriously_i_paid_80_to_have_vader_locked/dpqeaiw/?utm_source=reddit&utm_medium=web2x&context=3):
![image](https://user-images.githubusercontent.com/52135169/107407663-ee181280-6ad7-11eb-8622-fedbccd09b7f.png)

These two comments above aren't autofocused. So users have to comb through to see which comment is the one in question, but also they may think these are the only comments for this Reddit post when it actually has almost 3k comments.

And there is no apparent pattern to where these comments show up. For example, it can be confusing with some higher-level child comments such as [this link](https://libredd.it/r/StarWarsBattlefront/comments/7cff0b/seriously_i_paid_80_to_have_vader_locked/dppum98/?context=3) to this comment:
![image](https://user-images.githubusercontent.com/52135169/107409042-b5793880-6ad9-11eb-83a5-c920a39474da.png) [Teddit](https://teddit.net/r/StarWarsBattlefront/comments/7cff0b/seriously_i_paid_80_to_have_vader_locked/dppum98/?context=3) has the same problem.

**Describe the solution you'd like**
Some way to autofocus on a comment. Maybe it can mean bigger font or box. [Nitter does this without JavaScript](https://github.com/zedeus/nitter/issues/333#issuecomment-770308853), I think they put a hashtag and some character after it at the end of the URL, looks like [this is how they do it](https://stackoverflow.com/a/55055391). [An example Twitter comment](https://nitter.cattube.org/vanillavigilant/status/1328432240246747137#m).

**Describe alternatives you've considered**
When I really need to know which comment is which, I just head over to Reddit, find out which comment is which, then go back to Libreddit.

**Additional context**
[...]

PR #115, opened 2021-02-11
Add "View all comments" and "Show parent comments" buttons when viewing a single thread. Closes #65

### LINKED ISSUE #65: Feature Request | Add "Show parent comments"button when viewing a comment
When you navigate to a comment directly, eg. coming from a user's profile, you cannot easily see the context of the comment.

Add a "Show parent comments" button. This button would append "?context=9999" to the URL (or a similar value).

**Additional context**
IIRC, the number of nested comments is now capped at 3 by libreddit. To display a full context, deeper nesting should be supported.

robrobinbin:
Also includes improvements to the "More replies" button logic.

When viewing a comment, it is highlighted, addressing some of the concerns of #112.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -157,12 +157,12 @@ async fn main() -> tide::Result<()> {
 	// Browse user profile
 	app.at("/u/:name/").get(user::profile);
 	app.at("/u/:name/comments/:id/:title/").get(post::item);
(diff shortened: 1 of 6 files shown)


spikecodes:
Awesome, thanks!

PR #121, opened 2021-02-16
Add more rich meta tags

EnderDev:
I have added more metatags with the `ogp:` and `twitter:` prefix. 

I noticed that images weren't being shown in the embeds shown on Discord or Twitter.

You might have to add `https://libreddit.it` for the `:image` and `:url` metatags as it is just the relative path and the sites might not be able to locate it.


spikecodes:
Thanks! Looks great!

spikecodes (commented):
Looks good

PR #134, opened 2021-02-23
Added multi-staged container build and a user

pred2k:
This reduces the image size from 2Gb to 92Mb by only put the necessary files into the container.
Container now runs without root priviliges. [see why](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-2-set-a-user)

I hope all dependencies are added.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,4 +12,6 @@ COPY --from=builder /usr/local/cargo/bin/libreddit /usr/local/bin/libreddit
 RUN useradd --system --user-group --home-dir /nonexistent --no-create-home --shell /usr/sbin/nologin libreddit
 USER libreddit
 
+EXPOSE 8080
+
 CMD ["libreddit"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,15 @@
-FROM rust:latest
+FROM rust:latest as builder
 
 WORKDIR /usr/src/libreddit
 COPY . .
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,8 +1,15 @@
...
 RUN cargo install --path .
 
+
+FROM debian:buster-slim
+
+RUN apt-get update && apt-get install -y libcurl4 && rm -rf /var/lib/apt/lists/*
+COPY --from=builder /usr/local/cargo/bin/libreddit /usr/local/bin/libreddit
+RUN useradd --system --user-group --home-dir /nonexistent --no-create-home --shell /usr/sbin/nologin libreddit
+USER libreddit
+
```
pred2k, line 14: Define Port https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#expose
```suggestion

EXPOSE 8080

```

spikecodes:
I attempted to build and run Libreddit in [Play with Docker](https://labs.play-with-docker.com/) using this `Dockerfile` and there were no logs unlike the existing `Dockerfile`, are you able to reproduce that issue?

spikecodes:
Nevermind, the issue didn't seem to be related to the Dockerfile. I just pushed 47d01a0 which fixed it.

spikecodes (approved):
Looks good, thanks!

PR #143, opened 2021-03-04
Update screenshot in README.md

Mennaruuk:
New screenshot for v0.3.1. Also 35% lighter!


spikecodes:
Great idea! I've took one at a slightly different resolution so that it can serve as the "GitHub Social Preview" as well, thank you!

PR #153, opened 2021-03-18
Update README.md Docker instructions

TheCultLeader666:
I updated the Docker install instructions so that the latest version is always used, even when an image had been pulled previously.


spikecodes:
Fantastic suggestion, thanks for this PR!

spikecodes (commented):
Looks good!

PR #176, opened 2021-04-05
Add opensearch support

Leopardus4:
Hi,

It would be awesome to have opensearch support implemented in Libreddit!

[Docs](https://developer.mozilla.org/en-US/docs/Web/OpenSearch) for reference.

This PR adds basic OS support. However, I didn't tested it locally, so please check that it works as intended (especially relative links).

What this PR does not add: Search suggestions. They would be also cool, but it's not my technical stack. To add it, generating simple JSON is needed: [docs](https://developer.mozilla.org/en-US/docs/Archive/Add-ons/Supporting_search_suggestions_in_search_plugins). I guess new template would be needed here, similar to [search.html](https://github.com/spikecodes/libreddit/blob/master/templates/search.html).

### DIFF
--- a/static/opensearch.xml
+++ b/static/opensearch.xml
@@ -1,0 +1,11 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/"
+                       xmlns:moz="http://www.mozilla.org/2006/browser/search/">
+  <ShortName>Search Libreddit</ShortName>
+  <Description>Search for whatever you want on Libreddit, awesome Reddit frontend</Description>
+  <InputEncoding>UTF-8</InputEncoding>
+  <Image width="32" height="32" type="image/x-icon">/favicon.ico</Image>
+  <Url type="text/html" template="/search">
+    <Param name="q" value="{searchTerms}"/>
+  </Url>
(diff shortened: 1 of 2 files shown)


spikecodes:
I love this idea, thank you! Tomorrow, I can make Libreddit resolve `/opensearch.xml` to the XML file and then this will be complete!

spikecodes:
Unfortunately, the `Url` parameter in the [OpenSearch specification](https://developer.mozilla.org/en-US/docs/Web/OpenSearch#opensearch_description_file) doesn't support local paths so I'd probably have to create a new command line argument to pass in the URL of the instance which would allow the instance to enable OpenSearch support.

Leopardus4:
Darn, I hoped it will be simpler. Yeah, passing host URL via command line is a good solution here. It looks like [invidious](https://docs.invidious.io/Installation.md#usage) & [nitter](https://github.com/zedeus/nitter/blob/master/nitter.conf#L7) do it this way. Another option I thought about was to get the `Host` http header from request in runtime (is this possible in Hyper?) - just sayin'

Scoder12:
+1 for host header

seychelles111:
bump

skydiver1980 (approved):

PR #181, opened 2021-04-06
Fix comments

bnainar:


bnainar:
Done 😁

spikecodes:
Thanks! :)

bnainar:
@spikecodes, I only changed .GitHub folder, then how did GitHub Actions failed. Pls look into that

spikecodes:
GitHub Actions fails for any PRs here because the Docker Build requires a Secret that only gets pushed to the Action when I initiate the a runner.

Issue #201, opened 2021-04-20
List subreddit's mods.


jtagcat:
## Describe the solution you'd like
On non-thread pages, show a collapsed section below the sidebar, listing the mods of the subreddit.

spikecodes:
Great idea, thanks! :grin:

jtagcat:
Please use the heart or +1 reaction on github instead of commenting. It comes up as a notification each time.

curlpipe:
I have a PR. It's not a separate collapsed section, instead I integrated it into the sidebar itself. But it does the job.

curlpipe:
@spikecodes Afaik, this issue can be closed :)

jtagcat:
gh doesn't recognize `This PR closes issue no. #201`

Use `Closes #201` instead.

spikecodes:
Just linked them so that should be fixed now.

PR #213, opened 2021-05-03
Add list of moderators to sidebar

curlpipe:
So I added a list of moderators in bullet points to the sidebar. Each item has a link to that users profile.

This PR closes issue no. #201 

image
image

I also added in a message for when the wiki isn't found. This also fixes a UI rendering bug that makes everything bunch together in a ghastly fashion.
image


spikecodes:
This looks great, thank you! :)

PR #219, opened 2021-05-10
Added support for quarantined subreddits

### LINKED ISSUE #208: Support viewing quarantined subreddits
Quarantined Subreddits Cant Viewable like (example : r/fullcommunism)

Support Quarantined Subreddits

in old.reddit.com some quarantined subreddits viewable without login/email verification

curlpipe:
Tricked Reddit into thinking that the ~~user~~ instance has opted in to viewing quarantined subreddits.

Closes #208

- [x] Find method to bypass quarantine block.
- [x] Implement cookie boolean in functions
- [x] Write wall template for quarantine
- [x] Add cookie
- [x] Ensure all routes work
    - [x] /r/sub
    - [x] /r/sub/subscribe
    - [x] /r/sub/unsubscribe
    - [x] /r/sub/comments/*
    - [x] /r/sub/search
    - [x] /r/sub/wiki/*
    - [x] /r/sub/about/sidebar
- [x] Fix feed bug
- [x] Add quarantine wall to posts
- [x] Fix case issue

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -75,6 +75,7 @@ fn request(url: String) -> Boxed<Result<Response<Body>, String>> {
 		.header("Accept", "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8")
 		.header("Accept-Language", "en-US,en;q=0.5")
 		.header("Connection", "keep-alive")
+		.header("Cookie", "_options=%7B%22pref_quarantine_optin%22%3A%20true%7D")
 		.body(Body::empty());
 
 	async move {
--- a/src/client.rs
+++ b/src/client.rs
@@ -60,7 +60,7 @@ async fn stream(url: &str, req: &Request<Body>) -> Result<Response<Body>, String
 		.map_err(|e| e.to_string())
 }
 
-fn request(url: String) -> Boxed<Result<Response<Body>, String>> {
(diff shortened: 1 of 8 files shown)


curlpipe:
I have several questions

1. Should this be explicitly enabled by an option in preferences?
2. Should I not send the cookie header with each request, only sending it when there is quarantined Subreddit. As far as I can see, this will require 2 requests instead of just one, one to see if it's quarantined, another to actually retrieve the content.

I like this current implementation because it's quite simple and doesn't really slow anything down. Not sure if it will have any negative effects on privacy. :thinking:

spikecodes:
Thanks for opening this PR! To answer your questions,

1. I believe the best option would be following Reddit's model of prompting users to a dialog when they open a quarantined subreddit. The dialog would ask whether they want to proceed and pressing "Yes" would reload the subreddit with the `Cookie: _options=%7B%22pref_quarantine_optin%22%3A%20true%7D` header sent to Reddit (Libreddit should present no dialog in this case). I don't feel this should be saved to settings and shouldn't take up another cookie as the `pref_quarantine_optin` option would be enabled just for that single request.
2. Does my first response answer this question?

curlpipe:
Dang it! Conflicts. I'll fix them.

spikecodes (changes requested):
I made a few changes to the `quarantine` and `param` functions. Let me know if these look good to you and I'll merge!

curlpipe:
Looks all good to me

PR #220, opened 2021-05-16
Fix "HLS" typo in "Enable HLS" notification

invakid404:
This isn't anything important, but I noticed that "HLS" was misspelled as "HSL" in the notification which prompts you to enable it, so I thought I might as well fix it real quick :^)


spikecodes:
Good catch, thanks! :smile:

PR #223, opened 2021-05-22
Instances in computer readable format

RiversideRocks:
I've added the list of instances in CSV format. This might be useful in the future if someone wanted to create an app that could use the list of instances.


spikecodes:
Thanks! This suggestion has finally been fulfilled with the new [libreddit-instances](https://github.com/libreddit/libreddit-instances) repository!

PR #227, opened 2021-05-23
Include SystemD configuration

somini:
Added some documentation on the README, but you might want to review that.

I can also send a patch to AUR to integrate this on that package.


spikecodes (approved):
This looks great, thank you for contributing to Libreddit!

PR #237, opened 2021-07-01
Added rosebox theme

x6r:
i have been locally using this theme with [Stylus](https://github.com/openstyles/stylus) for quite a while so i  thought about officially adding it to the project instead

it is based on the [rosebox](https://github.com/KraXen72/rosebox) color palette and looks like this

preview

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -141,6 +141,20 @@
 	--shadow: 0 2px 5px rgba(0, 0, 0, 0.5);
 }
 
+/* Rosebox theme setting */
+.rosebox {
+	--accent: #a57562;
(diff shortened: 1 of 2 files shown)


spikecodes:
Thank you for the theme! I'm concerned that Libreddit has too many themes right now and that the dropdown has become quite cluttered. I want to either provide an alternative theme selection in the future (perhaps a grid?) or narrow down what themes can be accepted. IHopefully you don't mind but until then, I'll just let this sit here.

spikecodes:
Sorry, I don't think I'll ever add a theme browser so I'll just merge this.

PR #247, opened 2021-07-19
Encode html characters in flairs

accountForIssues:
[On Reddit](https://old.reddit.com/r/likeus/)

image

# Before

[Libreddit](https://libreddit.kavin.rocks/r/likeus)

image

# After

image

---
I am VERY new to rust so I apologize if something was wrong but I'm here to learn. Feel free to reject and make a proper PR :)


spikecodes:
Thanks for catching this! :)

Libreddit currently has the `esc!` macro to escape HTML characters in strings so I slightly modified this PR to use that macro instead of a new crate.

PR #275, opened 2021-09-09
Use proper spelling of `systemd`

lnicola:
Spelling it as SystemD tends to be a shibboleth among people who avoid it, so let's use the official spelling.


spikecodes:
Good correction! I'm not sure why I had the notion that it was spelled with a capital d, maybe I've seen it misspelled too many times.  Either way, I appreciate the fix.

PR #278, opened 2021-09-11
Add specific permissions to workflows under .github/workflows

varunsh-coder:
This PR adds specific permissions to the existing workflows under .github/workflows.

### Background

GitHub provides a [feature](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) to set permissions for the GITHUB_TOKEN. I have implemented a [GitHub App](https://github.com/apps/step-security) to automatically calculate the right permissions for a given workflow.  

I am trying the App out on public repositories, by forking them, installing the App on the fork, and manually creating PRs with the fixed workflows.  

I have manually reviewed the changes, and they do look good to me. If something looks off, please let me know. Please try it out on other repos and share your feedback. Thanks!

### DIFF
--- a/.github/workflows/docker-arm.yml
+++ b/.github/workflows/docker-arm.yml
@@ -9,6 +9,8 @@ on:
 
 jobs:
   build-docker:
+    permissions:
+      contents: read
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/docker-armv7.yml
+++ b/.github/workflows/docker-armv7.yml
@@ -9,6 +9,8 @@ on:
 
 jobs:
   build-docker:
+    permissions:
+      contents: read
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -9,6 +9,8 @@ on:
 
 jobs:
   build-docker:
+    permissions:
+      contents: read
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,6 +11,8 @@ env:
 
 jobs:
   build:
+    permissions:
+      contents: write
     runs-on: ubuntu-18.04
 
     steps:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,7 +12,7 @@ env:
 jobs:
   build:
     permissions:
-      contents: write
+      contents: read
     runs-on: ubuntu-18.04
 
     steps:


spikecodes:
Thanks for the suggestion! What do you mean by "the GITHUB_TOKEN"? Do you mean the `RELEASE_TOKEN` secret that's used by the [Rust workflow](https://github.com/spikecodes/libreddit/actions/workflows/rust.yml)? If so, none of the docker-related workflow configs use that so they shouldn't need editing. Where can I find the source code for your GitHub app?

varunsh-coder:
> Thanks for the suggestion! What do you mean by "the GITHUB_TOKEN"? Do you mean the `RELEASE_TOKEN` secret that's used by the [Rust workflow](https://github.com/spikecodes/libreddit/actions/workflows/rust.yml)? If so, none of the docker-related workflow configs use that so they shouldn't need editing. Where can I find the source code for your GitHub app?

Hey @spikecodes, all GitHub workflows get a GITHUB_TOKEN. There is more info on it [here](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) and [here](https://docs.github.com/en/actions/reference/authentication-in-a-workflow). By default, it has permissions to most of the GitHub APIs. If you look at any of the workflow runs in your repo, e.g. [this one](https://github.com/spikecodes/libreddit/runs/3576361025?check_suite_focus=true#step:1:13), you can see the permissions assigned to the GITHUB_TOKEN. 

The changes I am proposing will reduce the permissions of the token. All the workflows in your repo only need `contents:read` permissions, since `actions/checkout` [uses the token](https://github.com/actions/checkout/blob/afe4af09a72596f47d806ee5f8b2674ec07fdc73/action.yml#L24) to fetch code. 

`softprops/action-gh-release` action [uses the token](https://github.com/softprops/action-gh-release/blob/a3f0173fb37355ab0a460e5790c1229e010a4e87/action.yml#L36) from the action input (`with:`) or the env variable. It [recommends](https://github.com/softprops/action-gh-release#environment-variables) not using env variables in the future though... 

I can see now that the App's algo got confused because in your case token is specified in the env variable. Since you use the `RELEASE_TOKEN` in that action, the GITHUB_TOKEN should not need `contents:write` for that job either. I can manually modify permissions for the job in the PR to `contents:read`. Will fix the App soon to handle cases where token can be in action input or env variable...

Hope that clarifies. Once the `permissions` keys are added, the permissions [in workflow runs](https://github.com/spikecodes/libreddit/runs/3576361025?check_suite_focus=true#step:1:13) would show `contents:read` only. 

Please let me know if you have more questions. Unfortunately, the source code of the App is not open-source.

spikecodes:
Thank you for clarifing! That's a shame to hear that the App is not open source. Sorry but I wouldn't be willing to use it in Libreddit repository if Step Security remains closed source. I appreciate the suggestion though and wish you the best of luck on your GitHub App!

varunsh-coder:
Thanks @spikecodes for the feedback! Because of your feedback, I will seriously consider open sourcing it in the future.

PR #280, opened 2021-09-12
Added heroku.yml

dacousb:


spikecodes:
Are you able to provide a brief description as to what this pull request would do? Also, was there a specific reason for setting all the default settings to `false` in `app.json`?

dacousb:
- This small change will make deploying to Heroku much easier (https://devcenter.heroku.com/articles/heroku-button).
- The options are not set to false by default, but I have marked them as not required.

dacousb:
There seems to be some problems with port bindings in Heroku, I am investigating the problem. One solution would be to have the port work as follows (thing that should work with Heroku):
```rust
let port = env::var("PORT")
        .unwrap_or_else(|_| "8080".to_string())
        .parse()
        .expect("PORT is not a number");
```
or similar (this can also be handy for other platforms).

dacousb:
I fixed the port issue, now it works perfectly in Heroku (I tested it).

spikecodes:
Awesome, thanks for this pull request!

PR #282, opened 2021-09-16
Add cursor: pointer property to post summary bars

modulated:
Adds cursor: pointer CSS property on hover of summary bar to better indicate to user that posts can be clicked to collapse comment chain.


spikecodes:
Great idea, thanks!

PR #284, opened 2021-09-27
Update README.md

obeho:
Made a change for an instance's Cloudflare category. I'd been using their instance since so long but recently realized it's served through CF. So shifted to another instance.

A status column should be added to this instance list indicating a)their Libreddit version usage and b)up/down status. Several (listed below) are offline/defunct. Maybe like [this](https://github.com/zedeus/nitter/wiki/Instances) or better?

`libreddit.dothq.co`, `libreddit.40two.app`, `reddit.phii.me`, `libreddit.database.red`, `libreddit.sugoma.tk`, `libreddit.awesomehub.io`


SergioFLS:
https://libreddit.silkky.cloud/cdn-cgi/l/email-protection does indeed give the email protection screen. 😞
 Also a defunct instance column would be nice, at least for me

spikecodes:
Good catch, thanks! I appreciate the suggestion about a status column, would you be able to open a separate issue for that?

PR #309, opened 2021-11-01
Change healthcheck script to utilise "wget"

Chengings:
Current healthcheck script in docker-compose.yml has an error cause by missing "curl".

This PR replaced "curl" with "wget" since Dockerfile(s) already use "wget --spider --q".

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,6 @@ services:
     ports:
       - 8080:8080
     healthcheck:
-      test: ["CMD", "wget", "--spider", "-q", "http://localhost:8080/settings"]
+      test: ["CMD", "wget", "--spider", "-q", "--tries=1", "http://localhost:8080/settings"]
       interval: 5m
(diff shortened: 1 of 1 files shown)


TheFrenchGhosty:
@Chengings You should add the same arguments used by Teddit: `"--no-verbose", "--tries=1",`

That's what I use on my instance: https://github.com/PussTheCat-org/PussTheCat.org-Configs/blob/master/Services/Libreddit/docker-compose.yml#L19

Chengings:
@TheFrenchGhosty Thank for suggestion. I added '--tries=1'. But I still prefer '-q' option instead of '--no-verbose' since all Dockerfile(s) use this quiet option.

I also tested with invalid port (:9999) and route (http://localhost:8080/this-path-doesnt-exist), the '-q' option still display an error message.

Screenshot 2021-11-10

spikecodes:
Thanks for this PR, looks good!

PR #316, opened 2021-11-12
Fix "Follow" button not changing to "Unfollow"

ClockVapor:
Currently, case sensitivity is broken when following a user. For example:

Notice how the URL contains "SPEZ" in all caps, and so does the "u/SPEZ" part on the right:
Screen Shot 2021-11-12 at 4 19 08 PM

You can change the case in the URL, and the sidebar will match it:
Screen Shot 2021-11-12 at 4 19 21 PM

If I try to follow this user, the "Follow" button doesn't change to "Unfollow". Going into the settings page reveals why:
Screen Shot 2021-11-12 at 4 23 01 PM

The case of the subscribed user and the case of the username in the URL don't match.

This PR fixes this by using the fetched username in the `User` struct instead of the provided name which can have arbitrary casing. With the fix applied:
Screen Shot 2021-11-12 at 5 51 04 PM

And after following:
Screen Shot 2021-11-12 at 5 51 17 PM

**Note:** This is only really an issue if you manually type in a user's URL with incorrect casing. If you click a link to a user's profile instead, the URL will have the correct casing, and so the user added to the subscriptions list will also have the correct casing.

### DIFF
--- a/src/user.rs
+++ b/src/user.rs
@@ -70,7 +70,7 @@ async fn user(name: &str) -> Result<User, String> {
 
 		// Parse the JSON output into a User struct
 		User {
-			name: name.to_string(),
+			name: res["data"]["name"].as_str().unwrap_or(name).to_owned(),
 			title: esc!(about("title")),
 			icon: format_url(&about("icon_img")),
 			karma: res["data"]["total_karma"].as_i64().unwrap_or(0),


spikecodes:
Thank you for this PR!

Issue #69, opened 2021-01-22
Feature Request | Filter by keywords across post (titles and comments,) subreddits, and users


somoso:
**Is your feature request related to a problem? Please describe.**
As an example, I see posts about bananas on libreddit, and I'm not a fan of seeing posts and individual subreddits. There may even be a user on reddit who is interested in talking endlessly about bananas. 

**Describe the solution you'd like**
I would like to have a filter set (through cookies) where I can filter out posts, subreddits and/or  users by my chosen keyword(s). In my case, it would initially be banana. Then in the future I may have a grudge against mangoes, which I'd like to filter out too.

Perhaps for users, I could tap a button next to their name and get a popup or an additional tap to confirm I would no longer like to hear from this user.

**Describe alternatives you've considered**
Closing my eyes and my mind to the posts about bananas and users who speak such heathen about bananas.

**Additional context**
Mobile app versions of Reddit (Apollo comes to mind) has this filtering locally:

![Image of Apollo with local filtering](https://i.postimg.cc/hjbQ8NHr/IMG-0177-Copy.png)

spikecodes:
Good idea, thanks!

sharkusmanch:
Would also be great to be able to filter by flair. Many reddit apps are missing this sadly

spikecodes:
Thanks! This is a good idea, I'll open a dedicated issue for that.

ClockVapor:
I've started working on this - will provide an update soon!

TheCultLeader666:
> I've started working on this - will provide an update soon!

Maybe open a draft PR?

ClockVapor:
@TheCultLeader666 I dig some digging, but I'm not seeing an option to do that anywhere.

TheCultLeader666:
> @TheCultLeader666 I dig some digging, but I'm not seeing an option to do that anywhere.

https://github.blog/2019-02-14-introducing-draft-pull-requests/

ClockVapor:
> > @TheCultLeader666 I dig some digging, but I'm not seeing an option to do that anywhere.
> 
> https://github.blog/2019-02-14-introducing-draft-pull-requests/

Got it - thanks!

PR #317, opened 2021-11-13
Filter subreddits and users

ClockVapor:
Pull request for #69

Currently functioning:
* Filter/unfilter subreddits
    * Subreddits now have a "Filter" button next to the "Subscribe" button.
    * Filtered subreddits show in the settings menu below subscriptions.
    * When viewing a list of posts (subreddit feed, user profile, search), posts from filtered subreddits won't show up.
    * If you visit a filtered subreddit (or multireddit), a message will show indicating that the subreddit is filtered.
* Filter/unfilter users
    * User profiles now have a "Filter" button next to the "Follow" button.
    * Filtered users show in the settings menu below subscriptions.
    * When viewing a list of posts (subreddit feed, user profile, search), posts from filtered users won't show up.
    * When viewing a post, comments from filtered users will be replaced with a _(Filtered content)_ message.
    * If you visit a filtered user's profile, a message will show indicating that the user is filtered.
* Misc.
    * When viewing a list of posts, if all posts are filtered, a message will show indicating that all posts on the current page have been filtered. Note that this case is different than if the listing itself is filtered - it's the difference between visiting a filtered sub and visiting a sub whose page 1 contents are only posts written by a filtered user.
    * Subscriptions and filters in the settings menu are now links.

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -65,7 +65,7 @@ pub async fn find(req: Request<Body>) -> Result<Response<Body>, String> {
 
 	let url = String::from(req.uri().path_and_query().map_or("", |val| val.as_str()));
 
-	match Post::fetch(&path, String::new(), quarantined).await {
+	match Post::fetch(&path, quarantined).await {
 		Ok((posts, after)) => template(SearchTemplate {
 			posts,
 			subreddits,
--- a/src/search.rs
+++ b/src/search.rs
@@ -37,9 +37,11 @@ struct SearchTemplate {
 	params: SearchParams,
 	prefs: Preferences,
 	url: String,
+	/// Whether the subreddit itself is filtered.
+	is_filtered: bool,
 	/// Whether all fetched posts are filtered (to differentiate between no posts fetched in the first place,
 	/// and all fetched posts being filtered).
-	is_filtered: bool,
+	all_posts_filtered: bool,
 }
 
 // SERVICES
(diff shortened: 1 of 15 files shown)


ClockVapor:
So, implementing flair filtering is proving to be a bit more complex than filtering subreddits and users (each filter is a keyword + mode). I'd still like to implement it, but at this point it's just holding back the filtering that I've already completed. @spikecodes what do you think of integrating just subreddit and user filtering for now?

spikecodes:
I really appreciate all the effort you've put into this PR and I'd be happy to merge just the subreddit & user filtering that you've done so far. Would it be possible to strip out the reformatting/refactoring/renaming to just the new filters code? That would make it much easier to review and test, thank you.

ClockVapor:
Sure! I'll clean up the PR

src/utils.rs
```diff
@@ -219,7 +219,7 @@ pub struct Post {
 
 impl Post {
 	// Fetch posts of a user or subreddit and return a vector of posts and the "after" value
-	pub async fn fetch(path: &str, fallback_title: String, quarantine: bool) -> Result<(Vec<Self>, String), String> {
+	pub async fn fetch(path: &str, quarantine: bool) -> Result<(Vec<Self>, String), String> {
```
ClockVapor, line 222: Removed the `fallback_title` hack for comments. Posts on reddit can never have an empty title, so if the title of a fetched item is empty, we know it's a comment.

src/search.rs
```diff
@@ -109,7 +149,7 @@ async fn search_subreddits(q: &str, typed: &str) -> Vec<Subreddit> {
 			let icon = subreddit["data"]["community_icon"].as_str().map_or_else(|| val(subreddit, "icon_img"), ToString::to_string);
 
 			Subreddit {
-				name: val(subreddit, "display_name_prefixed"),
+				name: val(subreddit, "display_name"),
```
ClockVapor, line 152: Using `display_name` instead (with a corresponding change in search.html) because the alt-text looked like `r/r/subname`

src/utils.rs
```diff
@@ -515,7 +539,7 @@ pub fn setting(req: &Request<Body>, name: &str) -> String {
 
 // Detect and redirect in the event of a random subreddit
 pub async fn catch_random(sub: &str, additional: &str) -> Result<Response<Body>, String> {
-	if (sub == "random" || sub == "randnsfw") && !sub.contains('+') {
+	if sub == "random" || sub == "randnsfw" {
```
ClockVapor, line 542: Removed unnecessary check

templates/search.html
```diff
@@ -41,7 +42,7 @@
 				<div class="search_subreddit_left">{% if subreddit.icon != "" %}<img loading="lazy" src="{{ subreddit.icon }}" alt="r/{{ subreddit.name }} icon">{% endif %}</div>
 				<div class="search_subreddit_right">
 					<p class="search_subreddit_header"> 
-						<span class="search_subreddit_name">{{ subreddit.name }}</span>
+						<span class="search_subreddit_name">r/{{ subreddit.name }}</span>
```
ClockVapor, line 45: Corresponding change for the `display_name` change in `search.rs`

templates/search.html
```diff
@@ -41,7 +42,7 @@
 				<div class="search_subreddit_left">{% if subreddit.icon != "" %}<img loading="lazy" src="{{ subreddit.icon }}" alt="r/{{ subreddit.name }} icon">{% endif %}</div>
```
ClockVapor, line 42: Here's where the alt text was `r/r/subname` before

spikecodes:
Looks great, thanks so much!

ClockVapor:
Thanks for cleaning it up and merging it!

PR #324, opened 2021-11-19
Format post and comment votes with a decimal place, like vanilla reddit does.

mikupls:
Before this change, a vote count of 1999 was displayed as 1k, which is a pretty big gap. The displayed count also differed from what Reddit does. Now, the behaviour is consistent.

Added some tests for format_num.

---

First time that I'm writing tests in Rust, using floats, etc. so please let me know if anything here seems wrong!

I'm not aware of an open bug that references this issue, but it has been on my todo list for some time. From personal testing, the change works in the default style/layout. For posts with 100k+ votes, the display is a bit border-line (literally), but it still looks okay-ish IMO.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -535,12 +535,14 @@ pub fn rewrite_urls(input_text: &str) -> String {
 	})
 }
 
-// Append `m` and `k` for millions and thousands respectively
+// Format vote count to a string that will be displayed.
+// Append `m` and `k` for millions and thousands respectively, and
+// round to the nearest tenth.
 pub fn format_num(num: i64) -> (String, String) {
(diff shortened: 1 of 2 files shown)


spikecodes:
Thank you! This is a really nice improvement that I never thought about adding.

I've added a tad more padding to post scores now that they will be wider (with the first decimal place).

Also, your Rust skills look great from what I've seen!

mikupls:
I started reading the Rust book thanks to the motivation of contributing to libreddit, so I got something good out of it as well :smile: 

Also, thanks @Mennaruuk for sharing your reactions, it's the only way I can know that other users like my contributions :)

Mennaruuk:
Of course! Great job, Libreddit is better because of contributors like you and spikecodes!

PR #328, opened 2021-11-19
Add text post previews.

### LINKED ISSUE #327: Feature Request | Support Reddit-style text post preview
When scrolling through some page, e.g. frontpage, or a specific subreddit, text-posts do not have a preview for the text (post) content.

Reddit shows a short preview of text posts (I think ~5 lines). This makes it easier to decide whether you want to open the thread or not.
Libreddit should consider also showing a text preview for text-posts.

---

As always, we could make this new display optional, though I do not see a strong reason for this.

I haven't checked yet how difficult this will be API-wise.

Example "text" post `https://www.reddit.com/r/AskUK/comments/qwpdjv/what_are_some_internet_phrases_that_can/`. However, the problem is only apparent when see these posts while scrolling through a subreddit. So this is not easily reproducible. Usually scrolling through the frontpage there are at least some text posts.

mikupls:
Things seems to work, and I cannot find anything I'm breaking with this change. 
However, this change was hard because I got a bit confused between the two places we seem to be parsing similar things (utils.rs, and post.rs). Or maybe I was just tired.

The `post_body` class was chosen because it seems like it makes the `grid-template` magic work that I understand almost nothing about.

The [truncate](https://djc.github.io/askama/filters.html#truncate) value was chosen "randomly". The value is big enough to fit most posts while not making subreddit pages unusable. The bigger concern I had was that the new text preview content seems to squish the sidebar a bit much. For testing see e.g. `r/offmychest`.

As always, if you can see a better way to accomplish this, let me know!

Fixes #327.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -251,11 +251,20 @@ impl Post {
 			// Determine the type of media along with the media URL
 			let (post_type, media, gallery) = Media::parse(data).await;
 
+			// selftext is set for text posts when browsing a (sub)reddit.
+			// Do NOT use selftext_html, because we truncate this content
+			// in the template code, and using selftext_html might result
+			// in truncated html.
+			let mut body = rewrite_urls(&val(post, "selftext"));
+			if body == "" {
+				body = rewrite_urls(&val(post, "body_html"))
+			}
+
 			posts.push(Self {
 				id: val(post, "id"),
 				title: esc!(if title.is_empty() { fallback_title.clone() } else { title }),
 				community: val(post, "subreddit"),
-				body: rewrite_urls(&val(post, "body_html")),
+				body,
 				author: Author {
 					name: val(post, "author"),
 					flair: Flair {
--- a/static/style.css
+++ b/static/style.css
@@ -722,6 +722,7 @@ a.search_subreddit:hover {
 
 .post_title {
 	font-size: 16px;
+	font-weight: 500;
 	line-height: 1.5;
 	margin: 5px 15px;
 	grid-area: post_title;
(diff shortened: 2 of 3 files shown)


spikecodes:
This is a great idea, thanks for this PR!

What do you think about a gradient mask, which would add even more parity to Reddit's redesign? I feel like it makes the text preview flow better with the post panels.

image

spikecodes:
Now that there is more text to keep track of, I've increased the font weight of the post titles so they stand out from the preview text.

mikupls:
Hmm, I think there are pros and cons to it. Obvious pro for me is that it is easier to see that there is more to the post (other than the "...", which get added by the `truncate` call).

The main con for me, and also why I dislike this behaviour on vanilla reddit, is that the gradient mask also gets applied to rather short posts. Browsing `r/offmychest` just now, there are lots of posts that fit in their entirety on the page, yet the mask makes it seem like there is more to the post.

I think in general it is an interesting question whether we should always thrive to make things as similar to vanilla reddit as possible, but to me it seems the rather obvious answer is no. So for me it comes down to whether you believe that the gradient will be something that the majority of libreddit users would like to see :)

I personally would opt towards no by default. If we could only apply it when the post is actually truncated, then I'm neutral. In the end, it will need some time to use the feature and then notice whether it feels annoying or not, so feel free to go ahead either way - and thanks for the quick review as always!

mikupls:
So TL;DR, if in doubt, go ahead and merge; if I actually feel like the gradient is making my libreddit experience worse, you will see a PR from me that makes the truncate logic and styling smarter :)

PR #329, opened 2021-11-21
Hide text post previews in Compact mode

rubenelshof:
Currently text previews (#328) are also shown in Compact mode.
On vanilla reddit text previews are not shown in what they call "Classic" mode which appears to be the same as Compact mode in libreddit.

With this minor change I added display none CSS to hide the post previews in Compact mode.


spikecodes:
Good catch, thanks!

Issue #320, opened 2021-11-15
Feature Request | A redirect page


thecashewtrader:
## Is your feature request related to a problem? Please describe.
A small issue pops up is when sharing links with other people, especially those without an extension like Privacy Redirect.

## Describe the solution you'd like
A redirect page like https://redirect.invidious.io/ could be very useful, as it would provide a single base URI that works for everyone.

## Describe alternatives you've considered
Generally, the two options are:
###  Link to reddit
Linking to reddit is undesirable for obvious reasons, including privacy.

### Link to the instance you are using
Linking to your instance is suboptimal because the person may be in another region of the world where a different instance would work better for him/her. You might also be self hosting libreddit on localhost, which would not work either way.

## Additional context
One improvement over https://redirect.invidious.io I would suggest is to potentially store the user's preferred instance, if any, in a cookie or otherwise, which would help in faster redirects.

PR #331, opened 2021-11-21
Clean out dead instances

RiversideRocks:


spikecodes:
Thanks for this PR! I'm not sure how many of these are dead in the long-term or temporarily offline so as a long-term solution, I will create a dedicated Libreddit instances/redirect page which will display a list of instances, act as a redirect gateway and display the online/offline status of each instance. That should cover #320, #223 and this.

I'll definitely update Jamie's instance address though!

mikupls:
I was actually working on a very similar PR :) Good to know what is coming next!

Mennaruuk:
@spikecodes Pretty urgent: one of the instances’ domain (igna.rocks) seems to be seized by the FBI. However, the Libreddit subdomain is still working. If I were you, I’d remove the domain from the list of public instances. Also, I think it’d benefit you greatly if you can put a disclaimer, something along the lines of “use these instances at your discretion” since you have no control over ones other than the two you own. Thanks a lot.

PR #334, opened 2021-11-21
Refactor Media parsing

mikupls:
introduce testdata directory with raw reddit API JSON for testing parsing functions.

refactor Media::parse for slightly more readability.

Add various test cases.

---

### This PR

Test post: `https://www.reddit.com/r/antiwork/comments/qybiul/some_worker_out_there_has_quit_in_the_coolest/`. When I saw this post on libreddit, it only showed me a `link` type post, and redirected me to vanilla reddit; a bad experience.

After I did all the work in this PR to get around it, I realized that the "proper" solution here might have been to implement cross(?)-posts "correctly" instead. However, I believe that might take much more work, and this PR could be an okay intermediate solution.

If you do not think so, I'm also okay with it, and can go back to try and implement cross-posts correctly.

### Testing

I hope this testing pattern goes into the right direction. As a new contributor I'm a bit slowed down when refactoring code because I worry that I might break unrelated functionality, and starting to add tests gives me a bit confidence that things will be okay. I think there is no good way around having raw reddit API data for having proper tests.

I was a bit unsure if it will be okay to have raw reddit response data in the repo, but it seems like a pattern also in other popular repos, such as https://github.com/praw-dev/praw/blob/master/tests/integration/cassettes/TestDomainListing.test_hot.json#L75


src/utils.rs
```diff
@@ -512,7 +526,7 @@ pub fn format_url(url: &str) -> String {
 
 			match domain {
 				"v.redd.it" => chain!(
-					capture(r"https://v\.redd\.it/(.*)/DASH_([0-9]{2,4}(\.mp4|$))", "/vid/", 2),
+					capture(r"https://v\.redd\.it/(.*)/DASH_([0-9]{2,4}(\.mp4|$|\?source=fallback))", "/vid/", 2),
```
spikecodes, line 529: Does this `source=fallback` query parameter change the video?

spikecodes (changes requested):
I really like how you've refactored Media::parse! Eventually, I think it might be nice to embed the child post (from cross-posting) inside of the parent post but for now, I think just grabbing the media is fine.

Is there some other way to better document Libreddit's code so contributors are less daunted without test cases of this manner? I don't think Reddit API json should be pushed to the repository. Ideally, comments could be made in confusing areas of the code

src/utils.rs
mikupls: What do you mean by change? I think it is the same as not having the parameter, but the reddit JSON that I have been using gave me this parameter without giving me a `.mp4` extension, which initially made `format_url` return an empty string (which lead me to write the tests for it, and also pull the reddit JSON for testing)

spikecodes: That makes sense, thanks for catching that!

spikecodes:
Thanks for the clarification! Could the Reddit API JSON test data be pulled each test rather than stored in the repo?

mikupls:
It is certainly a possibility. But I believe that this would be a testing anti-pattern for many reasons: 
1. The data could change unexpectedly or temporarily from Reddit making the tests fail (e.g. site/api down). 
1. You suddenly cannot run tests anymore when you are offline. 
1. Every test run would be extremely slow: image in the future we have 10-20 urls for testing various cases, and a single test run would take 10-20 seconds for getting all that data from Reddit every time you do a change.
1. You force everyone that wants to run tests to connect to the reddit API.

I want to reiterate that I'm also unhappy with storing raw reddit data in the repo, but I don't know a good practical alternative that allows us to build good unit tests.

spikecodes:
I acknowledge your points but Reddit has an uptime of 99.93% and if we're sending many requests, they could be ran in parallel (Rust is great for asynchronous functions 😉) so the total time could be ~1 second. Could perhaps the testdata aspect of this PR be split into it's own PR? I'd be happy to merge your great refactoring of the Media::parse function along with your solution to cross-posted media.

mikupls:
Sorry, was pretty busy the last weeks. I will pick this up again; but if you want to go ahead sooner, feel free to cherry-pick my changes.

spikecodes:
Thanks again!

mikupls:
@spikecodes thanks for picking this up while I was gone!

Was there any reason you dropped the tests from utils.rs (`format_url_works`)? It did not use any reddit data. If it's okay with you, I will re-add them in a follow-up. (having tests for complicated regexes is usually a nice to understanding the intention)

PR #335, opened 2021-11-21
Render markdown correctly in text post previews by using selftext_html.

### LINKED ISSUE #332: Improve text post previews
Some additions to text preview are needed as follow-up to #328:
- I thought that using `selftext` (and not selftext_html) will properly display the newline characters, but seems like this is not the case. Maybe we can fix it with one of the template built-in filters.
- Markdown is not rendered in the preview, but it should be.

Comparison:

Screenshot from 2021-11-21 12-00-32

Screenshot from 2021-11-21 12-04-42

Meta: I don't think normal users have permissions to set/change labels on GitHub issues; I would have gone with "feature parity". Not sure if you want to allow this to anyone though because of possible spam?

mikupls:
I was mistakenly under the impression that we somehow render markdown ourselves, but turns out we just take whatever HTML reddit gives us, and we also need to do this for text previews.

Use CSS to limit the size of the previews instead of truncating in the template.

Fix table CSS.

---

Fixes #332.

For testing the table CSS, see e.g. `r/Cricket/comments/qytpct/match_thread_3rd_t20i_india_vs_new_zealand/`. I hope I'm not accidentally breaking anything else here.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -251,11 +251,8 @@ impl Post {
 			// Determine the type of media along with the media URL
 			let (post_type, media, gallery) = Media::parse(data).await;
 
-			// selftext is set for text posts when browsing a (sub)reddit.
-			// Do NOT use selftext_html, because we truncate this content
-			// in the template code, and using selftext_html might result
-			// in truncated html.
-			let mut body = rewrite_urls(&val(post, "selftext"));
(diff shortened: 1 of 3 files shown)


spikecodes:
Looks great, thank you!

PR #345, opened 2021-11-28
Set sub and user descriptions to `overflow-wrap: anywhere`

ClockVapor:
I noticed some subreddits have long descriptions with no word boundaries, which results in this:
Screen Shot 2021-11-27 at 5 58 01 PM

Using `overflow-wrap: anywhere` appears to fix this (disclaimer: I'm no CSS expert, so this may be the wrong way to fix it):
Screen Shot 2021-11-27 at 5 57 26 PM

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -365,6 +365,7 @@ aside {
 #user_description, #sub_description {
 	margin: 0 15px;
 	text-align: left;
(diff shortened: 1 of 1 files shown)


spikecodes:
Good catch, thanks!

Issue #348, opened 2021-11-29
HTML tags showing as text in search results


burrhole:
## Describe the bug
When using libreddit's search function, the text post search results include html tags

## To reproduce
Use the search bar on the main page or any subreddit and look at text posts

## Expected behavior
Some preview text should appear without HTML tags

## Additional context
Using the main instance at libredd.it

Attached is an image of what I'm referring to:
![image](https://user-images.githubusercontent.com/31171855/143793817-af62233b-db1d-45f9-bc02-a226a989bc73.png)

spikecodes:
Thanks for reporting this! Yeah, this appears to be the same issue as #196 so I'll close this issue in favor of that one.

Issue #196, opened 2021-04-15
Wiki: HTML source is not rendered.


jtagcat:
https://r.c7.ee/r/datahoarder/wiki/Software

spikecodes:
Thanks for this bug report! I'm trying to understand why this happens, I don't know how this wiki page specifically could be escaped. :thinking:

grandeljay:
Could the HTML comment Tag be of any relevance? `<!-- SC_OFF -->` Maybe it's not rendered because it's a text/comment and not HTML.

spikecodes:
I think it must be something to do with Libreddit's rendering method considering [the other pages](https://libredd.it/r/datahoarder/wiki/) also have that comment but the comment's tags, as you mentioned, aren't escaped in other pages like they are in this one.

spikecodes:
Happens in some search feeds too: https://libredd.it/r/rust/search?q=Vec&restrict_sr=on&sort=relevance&t=all

joshers:
Jinja automatically escapes HTML elements when rendering variables. If you wish to allow the raw HTML to be parsed normally, the "safe" selector needs to be included with the variable:

`{{ wiki | safe }}`

Now, this opens up the potential for nasties to sneak by, so there should probably be some escaping done on the backend for things like javascript.

joshers:
Ignore above, I mistakenly thought Jinja was being used (sorry, Python/Flask are what I'm most familiar with) and was behaving bizarrely.

As specifically related to this wiki problem, the issue is that the link on the main Datahoarder page has a capitalized "S", while the _actual_ wiki page does not. So when libreddit tries to grab the json for /wiki/Software, it gets redirected via 304 to /wiki/software. What is important here, is that the redirect does not preserve the GET variables, so raw_json is not set to one anymore, and the content_html that is returned is already escaped by the reddit API.

spikecodes:
@joshers Thanks so much for troubleshooting this!

PR #351, opened 2021-11-30
Fix for escaped HTML in some search returns

joshers:
I just appended the raw_json flag to the search GET request in order to return the unescaped content_html from the reddit API during searches. This is such a minor and straightforward change, I can't see it having an impact on any other features or functionality.


joshers:
This specifically addresses #348, which was merged with #196

spikecodes:
Thanks! Already covered by https://github.com/spikecodes/libreddit/commit/0656756d21899b99530a5eba15ed0cd6ef900c8f now

PR #360, opened 2021-12-10
Remove Cloudflare Proxy from libreddit.drivet.xyz

GhostSlayer:
I'm testing some stuff and as a result i have disabled proxy for libreddit.drivet.xyz. It exposes my public ip, but also gives more privacy i guess


GhostSlayer:
Turns out this change just broke everything 🤔 I have reverted my changed

GhostSlayer:
I installed Libreddit to another VPS, and decided to keep the dns records unproxied. So this pull request will be useful too :D

GhostSlayer:
Also the server is now hosted in Poland, if you can change that

PR #369, opened 2021-12-23
add new instance

EsmailELBoBDev2:
I want to add my own instance - i'm new to selfhosting so i can't guarantee it will be running all the time but i can guarantee whenever it's down i will fix it within hours!


spikecodes:
Awesome, thanks for hosting Libreddit!

Issue #383, opened 2022-01-04
Search in subreddit as default


kimchifridge:
I'd like to make a suggestion where the default search in a subreddit is to search within that sub rather than all of reddit. I'm guessing people would more often search within the sub when they're on one.

Issue #355, opened 2021-12-01
searching "r/<subreddit>" should automatically redirect to it


w1kee:
## Is your feature request related to a problem? Please describe.

When I write r/subreddit in the search box and press enter, it opens the search page and I have to click a second time to actually get to the subreddit.

## Describe the solution you'd like

Searching "r/subreddit" will automatically redirect to subreddit.

PR #384, opened 2022-01-05
Search fixes

andrew-kaufman:
Addresses #383 and #355


spikecodes:
Good additions, thanks!

PR #367, opened 2021-12-19
Don't create hrefs to u/[deleted]

alyaeanyx:
Fixes #363

PR #386, opened 2022-01-07
Fix and improve admin/mod distinguishers

674Y3r:
- ### Fix regression caused by #367

| Before | After |
|--------|-------|
|  |  |

- ### Add distinguishers for post authors

| Before | After |
|--------|-------|
|  |  |
|  |  |

**Note:** If reddit doesn't show the admin/mod badge, then this PR doesn't show it either (except for posts in search results that have it).


spikecodes:
Thanks for this PR!

Issue #389, opened 2022-01-09
can't copy images directly from the feed in Chromium


laalsaas:
## Describe the bug

When I'm right-clicking on an image in the libreddit UI with the chromium-based qutebrowser, I don't have the "Save Image/ Copy image/ Copy image URL" options. This is because chromium doesn't recognize the way libreddit currently represents images as html:
 ### Libreddit
 ```html

	<a href="/img/o7igrl4qkka81.jpg" class="post_media_image short" >
		<svg
			width="1536px" 
			height="2048px" 
			xmlns="http://www.w3.org/2000/svg">
				<image width="100%" height="100%" href="/img/o7igrl4qkka81.jpg"/>
				<desc>
					<img loading="lazy" alt="Post image" src="/img/o7igrl4qkka81.jpg"/>
				</desc>
		</svg>
	</a>
```
### Reddit:
```html
<img alt="Post image" class="_2_tDEnGMLxpM6uOa2kaDB3 ImageBox-image media-element _1XWObl-3b9tPy64oaG6fax" src="https://preview.redd.it/o7igrl4qkka81.jpg?width=640&amp;crop=smart&amp;auto=webp&amp;s=658b5c66052a0576bf3218ad5e619503cd4d4e59" style="max-height: 512px;">
```
See also (qutebrowser#6947](https://github.com/qutebrowser/qutebrowser/issues/6947)

## To reproduce

1. Go to 'libredd.it' with a Chromium-based browser
2. Right-click on an image
3. no "Copy Image" entry in the menu.

## Expected behavior
The images should be represented in the more popular <img> html tag,
so all chromium-based-browser users can copy images directly from the Feed
without having to click on each image to open it as a File.

mikupls:
Hey, I just found this issue after writing #390. It seems to fix this problem (at least in Chromium 96).

Before:
![Screenshot from 2022-01-09 23-48-46](https://user-images.githubusercontent.com/93015331/148704350-f4a5544f-6f23-4c43-a1e8-21689603e19a.png)

After:
![Screenshot from 2022-01-09 23-47-18](https://user-images.githubusercontent.com/93015331/148704321-a0e49c83-47db-43f5-951a-fa4eb0935006.png)

Hope this addresses your report!

Issue #246, opened 2021-07-19
Reddit hosted images not shown inline/proxied


midi:
## Describe the bug
Images hosted on:

- i.redd.it

- preview.redd.it

are shown as links, instead of shown embedded as with the original reddit site.

mimetism:
I'm just commenting to thank you for opening this issues regarding the bug I described earlier. I was not able to do so because I incurred in the "Too many requests" problem mentioned in issue #244 and thus could not try to reproduce the problem again.

mikupls:
+1, was just about to open an issue about this. I might try to work on this one next.

mikupls:
Good news is that the two example cases seem to be different from a reddit API perspective (so good that we can cover them both). Bad news is that it seems like for the i.redd.it case the reddit API does not give us a width/height so I'm unsure how to make this work.

In any case I'll need to touch utils.rs, which is currently blocked by #334, so waiting a bit.

PR #390, opened 2022-01-09
Display previews and inline images for reddit-hosted images.

mikupls:
Display previews and inline images for reddit-hosted images. Partially addresses #246. Also avoid division by 0 in the rendering code.

### DIFF
--- a/templates/utils.html
+++ b/templates/utils.html
@@ -94,8 +94,26 @@
 	</p>
 	<!-- POST MEDIA/THUMBNAIL -->
 	{% if (prefs.layout.is_empty() || prefs.layout == "card") && post.post_type == "image" %}
+
+	<!-- reddit-hosted images do not provide us with width/height information: the values will be 0. -->
+	<!-- To work around "content reflow" problems, use svg tags to keep the image in place. -->
+	<!-- This causes two problems: right-clicking images will have a very degraded menu: no "copy image (link)", and bad display of reddit-hosted previews. -->
 	<a href="{{ post.media.url }}" class="post_media_image {% if post.media.height < post.media.width*2 %}short{% endif %}" >
+		{% if post.media.height == 0 || post.media.width == 0 %}
+		<!-- debug: reddit-hosted: svg workaround image thumbnail -->
+		<svg
+			width="{{ post.media.width }}px" 
+			height="{{ post.media.height }}px" 
+			xmlns="http://www.w3.org/2000/svg">
+				<image width="100%" height="100%" href="{{ post.media.url }}"/>
+				<desc>
+					<img loading="lazy" alt="Post image" src="{{ post.media.url }}"/>
+				</desc>
+		</svg>
+		{% else %}
+		<!-- debug: simplified image thumbnail -->
 		<img loading="lazy" alt="Post image" src="{{ post.media.url }}"/>
+		{% endif %}
 	</a>
 	{% else if (prefs.layout.is_empty() || prefs.layout == "card") && post.post_type == "gif" %}
 	<video class="post_media_video short" src="{{ post.media.url }}" width="{{ post.media.width }}" height="{{ post.media.height }}" poster="{{ post.media.poster }}" preload="none" controls loop {% if prefs.autoplay_videos == "on" %}autoplay{% endif %}><a href={{ post.media.url }}>Video</a></video>
--- a/templates/utils.html
+++ b/templates/utils.html
@@ -94,16 +94,8 @@
 	</p>
 	<!-- POST MEDIA/THUMBNAIL -->
 	{% if (prefs.layout.is_empty() || prefs.layout == "card") && post.post_type == "image" %}
-	<a href="{{ post.media.url }}" class="post_media_image {% if post.media.height / post.media.width < 2 %}short{% endif %}" >
-		<svg
-			width="{{ post.media.width }}px" 
-			height="{{ post.media.height }}px" 
-			xmlns="http://www.w3.org/2000/svg">
-				<image width="100%" height="100%" href="{{ post.media.url }}"/>
-				<desc>
-					<img loading="lazy" alt="Post image" src="{{ post.media.url }}"/>
-				</desc>
-		</svg>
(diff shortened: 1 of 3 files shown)


mikupls:
Happy new year :wave: I'm back after a long vacation :) Great to see all the improvements to libreddit in the meantime.

---

The change in this PR might seem big, but I could not detect any breakage - I was not able to figure out why we have all these SVG elements in the first place. In any case, as described in #246, it seems that reddit does not give us image dimension data, so we have to be able to live without it. I hope this PR allows us to do so.

Diff when browsing:
Screenshot from 2022-01-09 23-11-30

Diff when inside the post:
Screenshot from 2022-01-09 23-12-06

(left is the new behaviour in both cases.)

mikupls:
I was just catching up on new issues, and this PR also seems to fix #389 (by removing the SVG tags).

ThePixelDeveloper:
Regarding the SVG changes: https://github.com/spikecodes/libreddit/pull/68

It seems the author wanted to have a non jumping page when the images failed to load.

spikecodes:
Welcome back @mikupls! :wave:

Thank you for this PR but the removal of the SVG wrapper would cause jumping (also known as [Content Reflow](https://developers.google.com/speed/docs/insights/browser-reflow)) during image loading or perhaps image failing, as mentioned by @ThePixelDeveloper. I can almost fix this but the lack of dimensions for Reddit-hosted images in the API prevents me from entirely preventing the jumping from happening.

What is strange is that `data["preview"]` is shown when viewing Reddit while logged in but is completely hidden when viewing while signed-out. This appears to only occur for i.redd.it (such as [here](https://www.reddit.com/r/EscapefromTarkov/comments/odn91o/therapist_level_1_now_sells_lucky_scav_junk_box/.json)) images not others. :thinking:

Logged in:
image

Logged out:
image

Therefore Libreddit cannot receive the dimensions and thus, can't load the Reddit-hosted images without causing content reflow. This wouldn't be an issue in individual posts as much as it would when viewing feeds that contain multiple Reddit-hosted images, causing reflowing on each load.

mikupls:
Thank you both for the explanations!

So what is the proposal or solution here? Only display these images in-line when inside the post? (I would prefer also having the previews, but I use my own instance, so I'm not really affected by the reflow)

If we keep the svg tags or use a similar work-around, please add a comment about the reflow considerations.

mikupls:
Hi :wave: 

I just rebased on master and added my old fix for backslash rewriting. I've been using both commits personally for some months now and did not encounter any problems myself.

spikecodes:
Hi mikupls! Just got back from my break and I'd love to accept this PR but content reflow still seems to be an issue. The SVG tags alone won't fix this as we still need to supply the image dimensions to make sure the place in held to prevent reflow. Would it be okay to only display these images in posts not feeds? The feed code could be commented out until we find a solution

templates/utils.html
```diff
@@ -104,6 +110,10 @@
 					<img loading="lazy" alt="Post image" src="{{ post.media.url }}"/>
 				</desc>
 		</svg>
+		{% else %}
+		<!-- debug: simplified image thumbnail -->
+		<img loading="lazy" alt="Post image" src="{{ post.media.url }}"/>
```
mikupls, line 115: Ah, I think we might have to set the width/height here explicitly in the path where we have non-zero values?

It hard for me to test content-reflow. If anyone knows how to do it reliably, let me know please.

spikecodes:
Sorry, I'm trying to understand your second work-around version. Are you thinking define a fixed space for previews without dimensions and then have the image fill vertically?

templates/utils.html
spikecodes: Yeah, if we have non-zero width and height, we should definitely use them.

To test content-reflow, I recommend refreshing with Control + F5, which clears cache so the image might take longer to load and you can see the split-second where the comments jump from the middle of the page to the bottom

mikupls:
So what the current version does is:
- if we have the dimensions just use a normal img without the svg (see #389 on why it breaks things)
- if we have no dimensions, use the svg as-is. **Except**: because we do not have dimensions we have to pick some size: I just picked "100%" so it will fill whatever space there already is. we could also make the space bigger or quadratic or so. There is just no good solution that will not look bad sometimes in our case: we do not have dimensions and we do not want to wait to have them (content reflow). The only other solution is not having previews at all. (That is my understanding, maybe someone can come up with something better still)

So in summary I think the answer to you question is yes.

templates/utils.html
```diff
@@ -94,7 +94,13 @@
...
-	<a href="{{ post.media.url }}" class="post_media_image {% if post.media.height / post.media.width < 2 %}short{% endif %}" >
+
+	<!-- reddit-hosted images do not provide us with width/height information: the values will be 0. -->
+	<!-- To work around "content reflow" problems, use svg tags to keep the image in place. -->
+	<!-- This causes two problems: right-clicking images will have a very degraded menu: no "copy image (link)", and bad display of reddit-hosted previews. -->
+	<a href="{{ post.media.url }}" class="post_media_image {% if post.media.height < post.media.width*2 %}short{% endif %}" >
+		{% if post.media.height == 0 || post.media.width == 0 %}
+		<!-- debug: reddit-hosted: svg workaround image thumbnail -->
 		<svg
 			width="{{ post.media.width }}px"
```
mikupls, line 105: Need to play around with what we can put here because we do not have width/height (they are 0 in this path)

mikupls:
Pushed another update which should now actually match the behaviour that I described in the comment above. Sorry for the confusion.

mikupls:
Well it would've been too nice if things worked... My inexperience with front-end is showing.

I already found a new problem, this time with css. In the current version we seem to miss at least a `height: auto;` which is currently provided by `.post_media_image svg` (which no longer applies in my version).

I apparently also tested the content reflow incorrectly; my new approach is to zoom out to the maximum on `r/pics` and reload the page without cache. Now I see that my current version still seems to have problems (maybe due to the missing background css that was provided by the svg before...)

I'll get some sleep and maybe revisit. If you like, we can also finish this PR without changes to `utils.html` for now.

Current broken example:
Screenshot from 2022-03-17 23-02-26

mikupls:
Closing in favor of #738.

Issue #81, opened 2021-01-31
Feature Request | Save posts


spikecodes:
**Is your feature request related to a problem? Please describe.**
None.

**Describe the solution you'd like**
A button on posts to save them to cookies for later viewing. The button should have an SVG icon (whether that be a star, bookmark, or something else) and send a POST request to save it to a cookie. Saved posts should be viewable at `/saved` like in Reddit.

**Describe alternatives you've considered**
None.

**Additional context**
This feature request was split from #23.

Loki-L1130:
@spikecodes Seems like you are busy. But would be great . If you merge this pull request. So, We might stop bookmarking Libreddit links.

PR #391, opened 2022-01-10
Add ability to save posts (Addresses #81)

hamiltondave:
- Addresses #81 
- Added a star SVG icon at the bottom of posts which will allow users to save or unsave it
- List of saved posts is stored in a cookie
- Saved posts can be viewed at /saved, which is linked in the feeds dropdown

libreddit-saved-posts


PAI5REECHO:
Nice to see this addressed!

Do saved posts work with:
> You can restore your current settings and subscriptions after clearing your cookies using [this link](#).

Also are you able to save individual comments, not just parent threads?

hamiltondave:
Thanks for taking a look at this. I just updated the branch so that saved posts can be restored with the link on the settings page. This just works for parent threads for now though.

hamiltondave:
This is rebased now and ready to test.

PAI5REECHO:
I gave this PR a try. There's some stuff I see an issue with:
- [ ] The save button when viewing an individual post needs to be aligned to the end of the flexbox (preferably after the upvote ratio so it'd be in the corner in the same place as `/saved`):
image
- [ ] It'd be nice to be able to search `/saved` if possible
- [ ] `saved_posts` should automatically span multiple cookies if one cookie is more than [4093 bytes](http://browsercookielimits.iain.guru/) to avoid https://github.com/spikecodes/libreddit/issues/352 from happening (something like `saved_posts1`, `saved_posts2`, `saved_posts3`, etc...) 
- [ ] Should be able save individual comments rather than entire posts for feature parity with using real reddit accounts. It's also beneficial because often people just want to be able to recount a single comment in a very large thread rather than the post content itself. Also sometimes comments can get deleted by moderators so they won't be findable otherwise.

hamiltondave:
Thanks for your feedback! I think this is all doable. I will work on these improvements when I have some time.

PAI5REECHO:
Another thing I noticed is that when you submit the `/saved/{ID}/save"` the redirect is just `?redirect=/` rather than `?redirect=/#{ID}` to resume your scroll position on the current page

Daniel-Valentine:
@hamiltondave: Now that Libreddit development has resumed, I'd like us to consider this for inclusion in the code base. Would you mind pulling the latest from master and resolving merge conflicts?

Keep in mind that with the merge of #637 the template block that generates the view for a post has been made a macro and moved into `templates/utils.html`, so any changes you've made to that block will need to be ported to that file.

skydiver1980 (approved):

PR #406, opened 2022-01-23
Restore post sorting preference by link

### LINKED ISSUE #405: settings restore link does not include default subreddit sort
If I change the default subreddit sort, and save it, the link at the bottom of the settings page, to restore my settings and subreddits, does not include that parameter. So using that link to restore, reverts to sorting by Hot.

I would expect this to be preserved along with the other settings.

ClockVapor:
Fixes #405


Scoder12:
LGTM

spikecodes:
Thanks for catching this!

PR #453, opened 2022-03-20
Harden Systemd configuration

erdnaxe:
These options increase the isolation of libreddit service.

I am not setting `CapabilityBoundingSet`and `AmbientCapabilities` to allow users to easily add `CAP_NET_BIND_SERVICE` to listen on ports less than 1024. I am not setting `PrivateUsers=yes` as a private user cannot have process capabilities on the host's user namespace and thus `CAP_NET_BIND_SERVICE` has no effect.

I have been running my instance with this configuration since August 2021 without any noticeable issue.

Thanks,

### DIFF
--- a/contrib/libreddit.service
+++ b/contrib/libreddit.service
@@ -11,5 +11,27 @@ Environment=PORT=8080
 EnvironmentFile=-/etc/libreddit.conf
 ExecStart=/usr/bin/libreddit -a ${ADDRESS} -p ${PORT}
 
+# Hardening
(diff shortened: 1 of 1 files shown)


spikecodes:
Thank you for this!

Does `RestrictNamespaces=yes` do anything? [The docs](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RestrictNamespaces=) seem to indicate that it takes a list of strings not a "yes" or "no" value

PR #462, opened 2022-03-22
Wrap long post urls.

mikupls:
Like https://github.com/spikecodes/libreddit/commit/59043456ba507658f24e387fe4cb6fa75c5a99e4 but for post_url.

Example of breakage:
x1

(source: `r/UkrainianConflict/comments/taa8l4/`)


spikecodes:
Thank you!

PR #466, opened 2022-03-26
Add keyboard shortcuts to PREV and NEXT links

tekdoc:
Add keyboard shortcuts to PREV and NEXT links using the accesskey global attribute. Underlined letters indicate the key. NB: different browsers use different keystroke combinations to activate access key shortcuts. For example, Firefox on Mac uses ⌃⌥ + accesskey.

### DIFF
--- a/templates/search.html
+++ b/templates/search.html
@@ -92,13 +92,13 @@
 			{% if params.before != "" %}
 			<a href="?q={{ params.q }}&restrict_sr={{ params.restrict_sr }}
 				&sort={{ params.sort }}&t={{ params.t }}
-				&before={{ params.before }}">PREV</a>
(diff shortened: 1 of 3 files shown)


mikupls:
I keep getting surprised with what is possible with raw HTML. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey lists some accessibility concerns though.

I still wonder if having some minimal javascript for basic features would be worth it instead.

tekdoc:
If avoiding javascript is an objective for libreddit, this gets the job done 😉

spikecodes:
Thanks for sharing this awesome HTML-only feature with me!

PR #467, opened 2022-03-27
build: enable LTO and strip the binary

tirz:
This PR enable the link time optimization and strip the binary to reduce his size in release mode.

With this PR, the build time of `cargo build --release --target x86_64-unknown-linux-musl` increased from 3m00s to 3m15s but the binary size dropped from 15M to 6.3M.

I do not have any metrics to see how effective is the LTO.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,3 +25,8 @@ tokio = { version = "1.21.2", features = ["full"] }
 time = "0.3.16"
 url = "2.3.1"
 rust-embed = { version = "6.4.2", features = ["include-exclude"] }
+
+[profile.release]
+codegen-units = 1
+lto = true
+strip = true


tirz:
Update: also set codegen-unit to 1.
The new size of the binary is 5.7M (tested with libreddit 0.22.9).

Their is still a remaining question... Setting `strip = true` also strip the symbols, which make the bugs harder to find in case of bug : the crash report is pretty hard to read.
Maybe setting `strip = "debuginfo"` (for a size of 6.4M) is good enough ?

tirz:
Conflicts resolved.

Daniel-Valentine:
Provided we don't see negative impacts to Libreddit response times, I'm for this. I do think it would be good to see the impact of the LTO, so let's do get metrics for those. We can start by evaluating the performance of test code.

Re:

> Their is still one remaining question... Setting `strip = true` also strip the symbols, which make the bugs harder to track.
> Maybe setting to `strip = "debuginfo"` (for a size of 6.4M) is good enough for a release profile ?

So long as the debug symbols remain in dev builds then I see no issue with stripping the release build's symbol table completely.

@spikecodes, @sigaloid: Thoughts?

spikecodes:
Thanks for this PR!

Regarding the reduced debugability of using `strip = true`, I don't think we'll be troubleshooting for code bugs using the release profile so I'm not worried about that.

@Daniel-Valentine On my machine (target = stable-x86_64-unknown-linux-gnu), I benchmarked the current release build to take **55s** and with the proposed changes, **102s**. I have a pretty powerful PC so I've noticed these numbers are greatly inflated on lower-end hardware.

Daniel-Valentine:
> @Daniel-Valentine On my machine (target = stable-x86_64-unknown-linux-gnu), I benchmarked the current release build to take **55s** and with the proposed changes, **102s**. I have a pretty powerful PC so I've noticed these numbers are greatly inflated on lower-end hardware.

I agree that the build time will take a hit. My concern though was about Libreddit's HTTP response times; if Libreddit takes longer to deliver an HTTP response, then I don't think the LTOs are worth it.

I'll do some local testing to see how what affect the LTOs have.

spikecodes:
Ah, I noticed a negligible difference in HTTP response times for the home page.

tirz:
@Daniel-Valentine as mentioned in the opening, I did not found a way to benchmark the LTO :/
But I am interesting by your try.

For information, here was my stupid try (spoiler: it has failed, but I just ran it today with Libreddit v0.25.0 to show you the numbers).


**LTO=false**

```sh
> cargo build --release
release [optimized] target(s) in 3m 39s


> wrk --threads 4 --connections 100 --duration 120s --timeout 5s --latency http://127.0.0.1:8080
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.12s   523.30ms   4.98s    91.05%
    Req/Sec    13.44      7.41    60.00     75.77%
  Latency Distribution
     50%    2.07s 
     75%    2.09s 
     90%    2.22s 
     99%    4.48s 
  5488 requests in 2.00m, 286.18MB read
  Socket errors: connect 0, read 0, write 0, timeout 66
Requests/sec:     45.72
Transfer/sec:      2.38MB
```


**LTO=true**

```sh
> cargo build --release
release [optimized] target(s) in 3m 48s


> wrk --threads 4 --connections 100 --duration 120s --timeout 5s --latency http://127.0.0.1:8080
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.02s   542.32ms   4.99s    91.07%
    Req/Sec    14.37      8.50    70.00     74.91%
  Latency Distribution
     50%    1.96s 
     75%    1.98s 
     90%    2.17s 
     99%    4.44s 
  5810 requests in 2.00m, 302.62MB read
  Socket errors: connect 0, read 0, write 0, timeout 50
Requests/sec:     48.40
Transfer/sec:      2.52MB
```

As you may see, the LTO made Libreddit slightly faster, but in fact this test benchmark my network and not Libreddit itself (spoiler: I am still stuck with ADSL) -_-'
The landing page have a bunch of image etc...
Also for some reason, most of the time was used to resolve the reddit.com DNS (I have no clue why). Adding Unbound as a DNS cache resolved this issue.

**LTO=true + DNSCache**
```sh
> #sudo apt install unbound


> wrk --threads 4 --connections 100 --duration 120s --timeout 5s --latency http://127.0.0.1:8080
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   275.85ms  500.81ms   4.22s    89.40%
    Req/Sec   253.97    121.31   821.00     73.80%
  Latency Distribution
     50%   85.13ms
     75%   89.42ms
     90%  865.84ms
     99%    2.43s 
  106969 requests in 2.00m, 551.22MB read
  Non-2xx or 3xx responses: 105920
Requests/sec:    891.12
Transfer/sec:      4.59MB
```

And if we want to try this with in a context of production, adding something like Varnish as a reverse-proxy cache should help to drop the latency near to 0ms on localhost.
Anyway, I guess it is pretty hard to benchmark Libreddit because it involve some **_network_**!!!

The real (but much longer) way to benchmark the latency should be to write our own unit bench tests while mocking the network calls when needed.

Daniel-Valentine:
Gave this some thought and I think we should bring this in, so I'm going to merge. Thanks for the contribution.

Daniel-Valentine (approved):

PR #471, opened 2022-03-30
Update post_body overflow-x behaviour to scroll.

### LINKED ISSUE #469: Content overflow
## Describe the bug
Content overflow with long tables.
image

## To reproduce
1. Go to post with long table (e.g. [this](https://libreddit.spike.codes/r/GlobalOffensive/comments/ts76qw/natus_vincere_vs_ago_esl_pro_league_season_15/)),
2. Scroll down,
3. See overflow.

## Expected behavior
No visible overflow.

## Additional context
On Reddit, such posts have horizontal scrolling.

mikupls:
Fixes #469.

x

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -893,6 +893,8 @@ a.search_subreddit:hover {
 	padding: 5px 15px 5px 12px;
 	grid-area: post_body;
 	width: calc(100% - 30px);
+	overflow-x: scroll;
+	scrollbar-width: none;
 }
 
 /* Used only for text post preview */


mikupls:
Actually this adds an ugly "empty" scrollbar to posts without any overflow as well (so all posts), which is bad.

x

I'll add `scrollbar-width: none;` which fixes this on Firefox, but is apparently not implemented elsewhere yet?

https://caniuse.com/?search=scrollbar-width
https://stackoverflow.com/questions/9664325/style-the-scrollbar-with-css-in-google-chrome-webkit

674Y3r:
Wouldn't it be better to add the overflow only to the overflowing elements instead of the entire post?
Something simillar to blockquotes on wiki pages.

https://user-images.githubusercontent.com/87250374/161390671-bea9882c-9452-4d44-b4b7-06343d34d7e5.mp4

mikupls:
I think I know what you mean, and that behaviour would be nice, but I imagine it would be quite hard to implement. 

AFAIK we are just taking whatever HTML reddit is giving us and throw it into the post_body (which, in this case, includes a `<table>`). With my very basic frontend knowledge, it seems like we might have to do some rewriting of the reddit HTML to accomplish what's going on in the video (and we want to mostly avoid that).

My understanding from the image below is that the `post_body` is the overflowing element, and adding the overflow css to e.g. the table instead won't fix the problem.

x

Of course if you have an idea how to accomplish it, I would be happy to see it.

spikecodes:
Thank you for this PR! I found a way to achieve what @674Y3r was talking about with only overflowing on the tables in https://github.com/spikecodes/libreddit/commit/47ab85710333de39da134515127c7b8deb94b062

mikupls:
Ah, nice! I tried this without adding the `max-width` I think. Feels good to learn more about frontend :)

PR #482, opened 2022-04-10
Blur NSFW posts

NKIPSC:
This PR adds an option to blur NSFW previews when browsing subreddits (it doesn't blur the post itself when opening it).

I'd also like to enable this only on the front page, it requires changing a bit more code so I didn't add it yet but I'd be happy to implement it if you want.


EdwardLangdon:
This is actually a really needed feature.

Issue #475, opened 2022-04-05
Intentional backslashes are removed in comments


spikecodes:
## Describe the bug
It appears #461 has had an unintended effect of removing backslashes that should be present in post comments.

## To reproduce

Compare https://libreddit.spike.codes/r/ProtonVPN/comments/qaxigk/opensuse_unofficial_guide_on_how_to_install/hl1iw6g/?context=0

![image](https://user-images.githubusercontent.com/19519553/161825862-e7235afb-052d-47a4-9aaf-de916295beda.png)

to https://reddit.com/r/ProtonVPN/comments/qaxigk/opensuse_unofficial_guide_on_how_to_install/hl1iw6g/?context=0:

![image](https://user-images.githubusercontent.com/19519553/161825820-7ba01986-bfa4-41eb-880e-01700323f702.png)

## Expected behavior
Backslashes that were typed by Reddit users should show on Libreddit.

mikupls:
Ah true, IIRC for some reason I thought that the function is only applied to URL (maybe because of the naming `rewrite_urls`), but a quick search shows that the entry body is usually passed to this. Therefore the function name (and comment) is also misleading at this point, because the function now does more than just rewriting urls.

Maybe we can somehow apply the backslash filtering only to things that look like URLs. 
In any case, adding more test cases will help for the future.

Let me know if you are working on this already, otherwise I can take a look.

mikupls:
Comparing the problem case from #281: https://www.reddit.com/r/NatureIsFuckingLit/comments/pfywdx/isistius_brasiliensis_cookiecutter_shark_mouth/hb7tusx.json with our new problem case: https://www.reddit.com/r/ProtonVPN/comments/qaxigk/opensuse_unofficial_guide_on_how_to_install/hl1iw6g.json

It seems like maybe we can get away with *only* filtering the url-encoded backslashes.

Issue #281, opened 2021-09-15
'\_' in URLs needs to be changed to '_'


microraptor:
## Describe the bug
Apparently new Reddit  changes `_` to `\_` in URLs. An when new Reddit or the app display the URL they change it back again. However libreddit doesn't do that yet (and neither old.reddit).

## To reproduce

See here:
https://libredd.it/r/NatureIsFuckingLit/comments/pfywdx/isistius_brasiliensis_cookiecutter_shark_mouth/hb7tusx
and here:
https://old.reddit.com/r/NatureIsFuckingLit/comments/pfywdx/isistius_brasiliensis_cookiecutter_shark_mouth/hb7tusx/
But if opened in new Reddit the URL is correct:
https://www.reddit.com/r/NatureIsFuckingLit/comments/pfywdx/isistius_brasiliensis_cookiecutter_shark_mouth/hb7tusx/

## Expected behavior
`\_` in Urls needs to be displayed as `_`

spikecodes:
Sorry, I'm trying to reproduce the error but I'm not sure if I understand what the problem is. The Libreddit and old.reddit.com links work fine for me, is that intentional? And aside from the domain, all three links have the same URL paths.

microraptor:
Sorry about the confusion, I was unclear in my description. I don't mean the Reddit links themselves, but the comment they link to.

In libreddit and old.reddit the comment reads:
`https://www.wikiwand.com/en/Cookiecutter\_shark`

In new Reddit and the app the comment shows the working link:
`https://www.wikiwand.com/en/Cookiecutter_shark`

mikupls:
+1, just encountered this for the first time myself: https://libredd.it/r/linux_gaming/comments/qwsc86/game_feels_choppy/hl8ap3v/?context=3

mikupls:
Okay, I think I got it. While we are replacing backslashes all over the place, we never remove the HTML-encoded version of it, which also shows up sometimes ("%5C"). I have a working patch ready for PR soon.

PR #487, opened 2022-04-16
Make rewrite_urls more precise with regard to backslash removal.

mikupls:
Adjust function description and add another test-case.

---

Addresses #475 and relates to #281.

So we have tests for both observed breakages now. However, now that I understand a bit more what is going on here, this solution is also far from perfect. It will start breaking again once someone intentionally puts `\_` into a comment AFAICT.

The intention of the function is probably to limit the scope of the regex replacement to <a></a> elements. However operating with regexes on HTML is known to be tricky (or "impossible"). I propose we live with this intermediary solution until someone finds another relevant breakage.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -605,14 +605,17 @@ pub fn format_url(url: &str) -> String {
 	}
 }
 
-// Rewrite Reddit links to Libreddit in body of text
+// Rewrite Reddit links to Libreddit in body of text.
+// Also fix superfluous backslashes and underscores in links.
 pub fn rewrite_urls(input_text: &str) -> String {
-
 	let text1 =
 		Regex::new(r#"href="(https|http|)://(www\.|old\.|np\.|amp\.|)(reddit\.com|redd\.it)/"#)
+			// Replace reddit-prefix in href with "/" (i.e. point to our local instance).
 			.map_or(String::new(), |re| re.replace_all(input_text, r#"href="/"#).to_string())
-			// Remove (html-encoded) "\" from URLs.
-			.replace("%5C", "").replace(r"\", "");
+			// Remove (html-encoded) "\" from URLs. These seem to occur inside the href-attribute only.
+			.replace("%5C", "")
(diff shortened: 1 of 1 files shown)


mikupls:
Added another commit to support "new.reddit.com"; discovered on `r/japan/comments/skbmu8/the_rjapan_basic_questions_thread_february_2022/`.

src/utils.rs
```diff
@@ -605,14 +605,17 @@ pub fn format_url(url: &str) -> String {
...
-		Regex::new(r#"href="(https|http|)://(www\.|old\.|np\.|amp\.|)(reddit\.com|redd\.it)/"#)
+		Regex::new(r#"href="(https|http|)://(www\.|old\.|np\.|amp\.|new\.)(reddit\.com|redd\.it)/"#)
+			// Replace reddit-prefix in href with "/" (i.e. point to our local instance).
 			.map_or(String::new(), |re| re.replace_all(input_text, r#"href="/"#).to_string())
-			// Remove (html-encoded) "\" from URLs.
-			.replace("%5C", "").replace(r"\", "");
+			// Remove (html-encoded) "\" from URLs. These seem to occur inside the href-attribute only.
+			.replace("%5C", "")
+			// Format underscores nicely for display.
+			.replace("\\_", "_");
```
spikecodes, line 618: What is the primary function of this line? Is this supposed to match `\_` rather than `_`?

spikecodes: Nevermind, I see you said this relates to #281.

mikupls:
Anything else that is needed here?

Daniel-Valentine:
Sorry this has been on the backburner. I'm going through outstanding issues and PRs and intend to get to this one soon. Bleiben Sie dran.

mikupls:
Thanks. Looks like this patch still applies cleanly. Don't expect me to be able to explain the regexes after a year though :)

skydiver1980 (approved):

PR #489, opened 2022-04-18
DRAFT: Embed css themes to simplify adding and testing new themes.

mikupls:
My idea here is that this change should simplify adding new themes for new contributors, and simplifies testing and adjusting themes for contributors and reviewers.

Currently we have to change 3 files (readme, settings.html, and the ever-growing style.css). I hope with this change we only need to drop in one new file and be done.

Note that during dev, rust-embed seems to read from disk, so "hot-reloading" changes to a theme file is possible.

I'm deep into "no idea what I'm doing here" with regard to some of the Rust code, so I'd be happy if anyone can show me how to make things cleaner.

List of TODOs: 
- [x] Remove duplicated ThemeAssets.
- [x] Resolve documentation in the readme.
- [ ] Consider how to avoid assembling the style.css content every read (though this is cached, so maybe not so bad?).

If you think this idea is not worth it, let me know :) I had my fun learning about rust-embed.


spikecodes:
I love this idea, thanks so much!

spikecodes:
> Remove duplicated ThemeAssets.

By this do you mean the use of the `ThemeAssets` struct in both main.rs and utils.rs?

mikupls:
Cool, glad to hear.

>By this do you mean the use of the ThemeAssets struct in both main.rs and utils.rs?

Yes, I wasn't able to figure out how to achieve this more cleanly (only using TheseAssets once). Do you have an idea?

spikecodes:
Ok that was pretty easy to do.

> Resolve documentation in the readme.

What did you mean by this?

mikupls:
Cool, that indeed looked pretty easy :)

>Did you mean updating [this wiki page](https://github.com/spikecodes/libreddit/wiki/Contribute-a-Theme)?

Yup, an update there, including showing contributors how to test new themes themselves. And then possibly a note on https://github.com/spikecodes/libreddit#change-default-settings that the values don't come from code in the case of themes, but from the files in the themes folder.

spikecodes:
Sounds good thanks! Regarding the README, I think it's fine to not include the note

PR #490, opened 2022-04-20
Added gruvbox-dark and gruvbox-light themes

MarshDeer:


MarshDeer:
Updated PR to fit the new themes system; should be functional now

README.md
```diff
@@ -188,7 +188,7 @@ Assign a default value for each setting by passing environment variables to Libr
 
 | Name                    | Possible values                                                                                     | Default value |
 |-------------------------|-----------------------------------------------------------------------------------------------------|---------------|
-| `THEME`                 | `["system", "light", "dark", "black", "dracula", "nord", "laserwave", "violet", "gold", "rosebox"]` | `system`      |
+| `THEME`                 | `["system", "light", "dark", "black", "dracula", "nord", "laserwave", "violet", "gold", "rosebox", "gruvboxdark", "gruvboxlight"]` | `system`      |
```
Daniel-Valentine, line 191: ping @spikecodes, @sigaloid 

As a part of the changes in this PR I'm fine with the overflow since it doesn't actually have an impact on how the markdown table is rendered. I think though we should reconsider our pretty alignment of the table in the markdown source as the valid values for environment values only get longer, which means other rows will also be adjusted, which also means annoying overflows in terminal editors of each row. But this is something we can tackle outside this PR.

Daniel-Valentine (approved):

README.md
MarshDeer: Yup, I noticed the alignment got broken by adding new values, but I figured it would render properly anyways and editing other lines so the source would be aligned felt more invasive than it was worth for this pretty minor PR.

PR #501, opened 2022-05-07
Add instance run on Flux

j9d3it:
[Here is the tweet announcing it.](https://twitter.com/RunOnFlux/status/1474049461596868614) It's pretty slow to load but it should be running for a while, unless they shut it down. It's running on a few nodes of the network at any given time, so availability is good.


spikecodes:
Thanks for submitting this! I'd be happy to list it if the devs gave permission

PR #508, opened 2022-05-12
Implement "posts hidden because of NSFW"

### LINKED ISSUE #159: Feature Request | When NSFW is disabled in settings, show age gate for NSFW subreddits
Newcomers might be confused when visiting a NSFW subreddit. Currently, Libreddit just shows the side panel, with no posts. It might give the impression the subreddit is empty. On Reddit, an age gate is shown, giving the user a choice to either proceed (yes) or return to previous page (no).

An age gate warning for those who have NSFW disabled in settings. This solution can feed two birds with one scone: on one hand, the user is shown an age gate. On the other hand, if the user says yes, then the yes button would lead to enabling NSFW in the settings and making it persist. This is how Reddit does it for non-signed-up users.
**Describe alternatives you've considered**
When I visit a new instance of Libreddit and check out a NSFW subreddit, I make sure to enable NSFW in settings so I can see the subreddit's content.

**Additional context**
This is how Libreddit currently handles NSFW subreddits when NSFW is disabled in settings.
imageedit_2_3465049298

sigaloid:
Fixes #159

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -42,6 +42,8 @@ struct SearchTemplate {
 	/// Whether all fetched posts are filtered (to differentiate between no posts fetched in the first place,
 	/// and all fetched posts being filtered).
 	all_posts_filtered: bool,
+	/// Whether any posts were hidden because they are NSFW (and user has disabled show NSFW)
+	any_posts_hidden_nsfw: bool,
 }
 
 // SERVICES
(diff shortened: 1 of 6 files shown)


spikecodes:
I like the idea of this but it seems to show up in pretty much ever feed under this current implementation, including r/popular:
image

It might be best to curtail this to only NSFW subreddits where no content is viewable without enabling NSFW.

sigaloid:
Makes sense! I'll make the change so that it becomes "all" posts hidden by NSFW rather than any.

sigaloid:
Fixed logic to cover only when *all* posts are NSFW and user has disabled NSFW.

Daniel-Valentine:
@sigaloid: I'm going to close this in favor of the code you submitted to Ferrit, which is now available in a separate branch [`all-posts-hidden-nsfw`](https://github.com/libreddit/libreddit/tree/all-posts-hidden-nsfw). I'll open up a PR for that shortly.

PR #509, opened 2022-05-15
Let native elements use theme accent colour

EnderDev:
This PR makes all unstyled native elements like HTML checkboxes and inputs follow the theme's accent colour. See before/after below.

Screenshot 2022-05-15 at 15 14 22

Screenshot 2022-05-15 at 15 14 05


spikecodes:
Thanks for this PR!

PR #511, opened 2022-05-17
Made instance url consistant

BobIsMyManager:
Without `http://`, it [may have caused problems](https://github.com/libredirect/libredirect/commit/28068798d2dc68cf123ad9750340e7fc30d279ef#r73869639) for [libredirect](https://github.com/libredirect/libredirect)


spikecodes:
Thanks for catching this!

PR #517, opened 2022-05-26
remove 40two.app - dead/serves ads

sybenx:
40two.app looks like it serves ads instead of libreddit. Hasn't worked for 1 week+


spikecodes:
Thanks!

PR #524, opened 2022-06-04
L: Fixed two swapped config variables in the documentation.

artemislena:
L: Uh, seems like you swapped the `POST_SORT` and `COMMENT_SORT` variables in the README, this PR should fix that.


spikecodes:
Thanks for catching this!

PR #537, opened 2022-06-21
Add ~vern onion instance

gi-yt:
This PR adds vern.cc's onion instance of libreddit to instance list


spikecodes:
Awesome, thanks!

PR #552, opened 2022-07-14
Speed up heroku deployment

supershadoe:
### Changelog
- Use `Dockerfile.fromimage` instead of the default `Dockerfile` to directly deploy the docker image from Dockerhub.
This makes deployment much faster than building it in heroku's containers.
- Add new env vars and fix some errors.
Some env vars in `app.json` missed `_DEFAULT_` and `AUTOPLAY_VIDEOS` env var was missing.
- Also, added default values for env vars.
_Can be removed in case it's deemed useless._


supershadoe:
Fixed merge conflicts and rebased the branch

PR #567, opened 2022-08-14
Use singular form of "comment" for posts with 1 comment

arthomnix:
Use the singular of "comment" instead of the plural in the footer of posts with only one comment (i.e. "1 comment" instead of "1 comments").


spikecodes:
Thanks!

PR #570, opened 2022-08-18
Remove lots of dead instances from the instance list

Dyras:
Hello! I tested almost all links in the instance list, and these are the ones that I found not to work

I believe you can close https://github.com/spikecodes/libreddit/pull/545 if you accept this pull request, as this removes all of those instances.

A suggestion for adding new instances would be to put them at the bottom. That way it will be more clear which instances stick around or not. Just a thought.

Edit: Or maybe that's already the case. If so never mind me!


Daniel-Valentine:
Thanks for your PR. We've fixed this with the move of our instances list to [libreddit/libreddit-instances](https://github.com/libreddit/libreddit-instances).

PR #591, opened 2022-09-11
Add new instance (libreddit.oxymagnesium.com)

OxyMagnesium:
New instance that I'm hosting. US-NY, no Cloudflare.


spikecodes:

PR #611, opened 2022-11-03
Add DoomOne theme, c/o Tildemaster <root@vern.cc>

Daniel-Valentine:
This was a theme that Arya, a participant in the Ferrit matrix room, contributed to the Ferrit project, although I made the contribution on their behalf.


sigaloid (approved):
Not sure how I feel about the contrast on some of the text in the top bar, but that's why it's an option. LGTM :+1:

Issue #186, opened 2021-04-09
Brotli text compression


spikecodes:
## Is your feature request related to a problem? Please describe.
N/A

## Describe the solution you'd like
Server responses should be compressed using the Brotli text compression algorithm. This could be achieved through [this crate](https://github.com/dropbox/rust-brotli).

## Describe alternatives you've considered
GZIP compression, though it is less efficient than Brotli.

## Additional context
Forked from #185.

FireMasterK:
Shouldn't this be done on a reverse proxy level? 🤔

On nginx, I use https://github.com/google/ngx_brotli

spikecodes:
Thanks for showing me that repo! I think you're right. Like with TLS, the method of text compression should probably fall on the reverse proxy. I will be closing this issue for now.

PR #612, opened 2022-11-04
HTTP compression (Reddit -> Libreddit -> client)

Daniel-Valentine:
(This is a port of ferritreader/ferrit#54, with some improvements. Notably, the compressed body is cached, so Libreddit is only compressing upon receipt of a new response from Reddit _or_ if the last response was 10 minutes old.)

Implements HTTP compression, between both Reddit and Libreddit and Libreddit and a web browser. Compression between Reddit and Libreddit is mandatory, whereas compression between Libreddit and a client is opt-in (client must specify a compressor in the Accept-Encoding header).

Supported compressors are gzip and brotli. gzip support is ubiquitous, whereas brotli is supported by almost all modern browsers except Safari (iOS, iPhone, macOS), although Safari may support brotli in the future.

This resolves ferritreader/ferrit#4 and #186. The benefits are self-evident. Here is the results of `GET /r/pics/comments/haucpf/ive_found_a_few_funny_memories_during_lockdown/`, one of the biggest threads ever hosted on Reddit, without compression and with. Under brotli, the payload is reduced by a factor of 10.8.

```
$ time curl -s localhost:8080/haucpf | wc -c
298722
curl -s localhost:8080/haucpf  0.01s user 0.00s system 0% cpu 11.517 total
wc -c  0.00s user 0.00s system 0% cpu 11.517 total

$ time curl -s -H "Accept-Encoding: br" localhost:8080/haucpf | wc -c
27647
curl -s -H "Accept-Encoding: br" localhost:8080/haucpf  0.01s user 0.02s system 0% cpu 6.337 total
wc -c  0.00s user 0.00s system 0% cpu 6.336 total

$ time curl -s -H "Accept-Encoding: br" localhost:8080/haucpf | wc -c
27647
curl -s -H "Accept-Encoding: br" localhost:8080/haucpf  0.02s user 0.01s system 1% cpu 1.675 total
wc -c  0.00s user 0.00s system 0% cpu 1.674 total
```

@spikecodes' did leave the below [comment](https://github.com/libreddit/libreddit/issues/186#issuecomment-844556398) in #186:

> Like with TLS, the method of text compression should probably fall on the reverse proxy.

Maybe we can improve this to support a server disabling compression on either part of the Reddit<->Libreddit<->client link, so that someone running an instance can disable compression if they don't need it (i.e. their reverse proxy will do it for them)?


sigaloid:
> Maybe we can improve this to support a server disabling compression on either part of the Reddit<->Libreddit<->client link, so that someone running an instance can disable compression if they don't need it (i.e. their reverse proxy will do it for them)?

I'm not sure of the behavior with reverse proxies - do they strip the client header requesting compression? Do they simply compress the response if the local backend doesn't compress it? In any case it shouldn't cause breakage. I agree the reverse proxy would ideally handle it but lots of people don't mess with that kind of stuff - front-end compression could definitely be made an option and possibly on by default. I think there's no issue with backend compression between Reddit though - that would reduce a lot of the size of some payloads which is great.

src/server.rs
```diff
@@ -1,17 +1,83 @@
...
+impl ToString for CompressionType {
+	fn to_string(&self) -> String {
+		let s: &str = match *self {
+			CompressionType::Gzip => "gzip",
+			CompressionType::Brotli => "br",
+
+			_ => "",
+		};
+
+		s.to_string()
```
sigaloid, line 77:
```suggestion
		match self {
			CompressionType::Gzip => "gzip".to_string(),
			CompressionType::Brotli => "br".to_string(),
			_ => String::new(),
		}
```

This can be simplified for readability's sake

Daniel-Valentine:
@sigaloid: I think it should be the responsibility of whomever sets up their reverse proxy makes sure it doesn't double-compress. But I did some research on nginx, [which is the most popular reverse proxy](https://en.wikipedia.org/wiki/Nginx#Popularity), and have learned that:

* nginx will not strip any client headers unless you tell it to do so
* [nginx appears to be smart enough to recognize compressed content to deliver and will not try to encode it again](https://docs.nginx.com/nginx/admin-guide/web-server/compression/) (emphasis mine): "Compressing responses often significantly reduces the size of transmitted data. However, since compression happens at runtime it can also add considerable processing overhead which can negatively affect performance. NGINX performs compression before sending responses to clients, **but does not “double compress” responses that are already compressed** (for example, by a proxied server)."

So if you're setting up Libreddit behind nginx, you have nothing else to do to ensure data is conveyed to a client appropriately.

src/server.rs
```diff
@@ -213,3 +274,480 @@ impl Server {
...
+			($x:expr) => {
+				$x.to_string().as_str()
+			};
+
+			($x:expr, $($y:expr),+) => {
+				format!("{}, {}", $x.to_string(), ae_gen!($($y),+)).as_str()
+			};
+		}
+
+		for accept_encoding in vec![
```
sigaloid, line 670:
```suggestion
		for accept_encoding in [
```

No need to create new vec, you can iterate over an array :crab:

sigaloid (approved):
Looks good!

Daniel-Valentine:
@sigaloid: Thanks for the approval and suggestions, since implemented.

@spikecodes: Any thoughts? Would like your input before this goes in.

spikecodes (approved):
Looks great, thanks!

PR #636, opened 2022-11-08
Blur NSFW posts.

Daniel-Valentine:
Reimplementation of libreddit/libreddit#482 as introduced into Ferrit.

Originally authored by @NKIPSC <15067635+NKIPSC@users.noreply.github.com>


sigaloid (approved):
An errant thought I had: Maybe turn this on by default. That way even if someone enables "show NSFW posts" it still won't display the actual image. This is very opinionated though, one person saying it shouldn't would change my mind. 

In all other ways this PR looks great to me, approved.

spikecodes (approved):
Great PR!

PR #644, opened 2022-11-21
make the column in posts consistent

laazyCmd:
The width isn't consistent in every post: likely because some posts have content that doesn't reach the `max-width: 750px`.

Without `width: 100%`:
image

With `width: 100%`:
image

Reference: https://libreddit.spike.codes/r/bleach/comments/yzb0o7/episode_6_got_me_like/

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -154,6 +154,7 @@ main {
 }
 
 #column_one {
+	width: 100%;
 	max-width: 750px;
 	border-radius: 5px;
 	overflow: inherit;


Daniel-Valentine:
I don't have an objection to this, although I would like @spikecodes and @sigaloid to weigh in.

sigaloid:
Nor do I. :smile::+1:

Daniel-Valentine:
Merged in 92f5286.

PR #648, opened 2022-11-27
Fix embeds

Macic-Dev:
This PR makes the embeds for Discord and Telegram prettier by displaying the post's video, if there is one, using OpenGraph meta tags, similarly to [Teddit](https://codeberg.org/teddit/teddit).

Before:
toQBkX

After:
HaVXRI


sigaloid (approved):
Looks great! Thank you for the contribution.

Daniel-Valentine (approved):

PR #656, opened 2022-12-05
Landing page for NSFW content, SFW-only mode

Daniel-Valentine:
This PR introduces, with some modifications, two features implemented in the Ferrit fork:

1. A landing page for NSFW content when a user has not elected to view such content. See the original PR for screenshots: ferritreader/ferrit#25
2. SFW-only mode. Original PR: ferritreader/ferrit#24

The behavior as implemented in (1) is unchanged -- see that PR for examples of what the landing pages look like.

(2) is implemented slightly differently:
1. Majority of the back-end code is identical, except that now the setting is controlled by the environment variable `LIBREDDIT_SFW_ONLY`. Moreover, the Libreddit will only operate in SFW-only mode if the value of `LIBREDDIT_SFW_ONLY` is `on`; this is done so that the setting of this environment variable is consistent with how other default user settings are specified as environment variables.
2. Instead of a briefcase emoji, the navbar Libreddit logo is changed to green to reflect that the instance is SFW-only. There will still be a hint that appears in hover-over that explains that the instance is SFW-only.
Screenshot 2022-12-04 at 17-19-01 The Rust Programming Language


spikecodes:
Thanks for this PR! I like the idea of adding instance settings, not just default user settings.

I'm not sure about the green look for the Libreddit logo. Even though it's color is dependent on the user theme, I felt like the teal-grey combo was part of the branding. I think the briefcase might have been better at signalling the SFW-only flag without affecting the logo. 

Either way, is it necessary to list in the nav bar (and thus, on every page) that the instance is SFW-only? We could include that information in the proposed instance info page and only mention it elsewhere when the user is attempting to access NSFW content.

sigaloid:
> Either way, is it necessary to list in the nav bar (and thus, on every page) that the instance is SFW-only? We could include that information in the proposed instance info page and only mention it elsewhere when the user is attempting to access NSFW content.

I personally originally included an example briefcase emoji when the instance is SFW-only to make it more obvious to users that the instance will not display NSFW content, but it doesn't necessarily need to be included. 

Though maybe a footer with some instance info could be added, that's a feature request for another day.

Daniel-Valentine:
@sigaloid, @spikecodes, and I discussed this privately and we've agreed not to do the change to the logo. Instead, there will be a notice in the footer indicating if the instance is SFW-only. This footer message will probably become a link to a page detailing the instance's info, including default settings specified in the environment/config, itself including whether or not an instance is SFW-only.

I'll follow up with new changes later in the week.

Daniel-Valentine:
@spikecodes, @sigaloid:

Sorry for the delay. I think this is now ready for a re-review. Addressing Spike's concern, the footer will now indicate if the Libreddit instance is SFW-only (this indication only appears if the instance is as such). This is probably good enough for now. When we start implementing an instance information page, we can change the footer to something else.

Screenshot 2022-12-31 at 21-02-26 The Rust Programming Language

sigaloid (approved):
This looks good to go in :+1: I have some thoughts about the setting examples in the readme but we'll resolve that when config goes in.

Shall we approve, or wait for Spike's re-approval?

spikecodes (approved):
Looks great, thanks!

Daniel-Valentine:
@spikecodes: Thanks for addressing the merge conflict ahead of me.

PureAnarchy:
Is this automatically enabled for sites with version 26.0 and higher? Or do I have to enable it somewhere?

sigaloid:
It's automatically enabled, it just makes it more clear when the user can't view NSFW content due to their settings.

Daniel-Valentine:
> Is this automatically enabled for sites with version 26.0 and higher? Or do I have to enable it somewhere?

If you host an instance, set `LIBREDDIT_SFW_ONLY=on` in the program's environment to turn this on. In other words:

* If you launch the binary, this suffices: `LIBREDDIT_SFW_ONLY=on libreddit`
* If you use the libreddit/libreddit docker image, this suffices: `docker run -p 80:8080 -e LIBREDDIT_SFW_ONLY=on libreddit/libreddit`

PR #659, opened 2022-12-08
Add comment count in post

### LINKED ISSUE #654: ✨ Feature parity: Display the number of comments in the post itself
## How does this feature work on Reddit?

On Reddit, viewing a post shows the number of comments in the post itself.
image

## Describe how this could be implemented into Libreddit

Display the number of comments a post contains within itself.

#42 (corresponding PR: #60) already sort of laid the groundwork: for a post that is displayed in a subreddit / user profile, the number of comments is already displayed under the title.
image

The (extremely cumbersome) alternative to viewing the number of comments in a post is to go to Reddit itself and check the number of comments there.

Right now, viewing a post does not show the number of comments said post contains:
image

sigaloid:
Fixes #654

### DIFF
--- a/templates/utils.html
+++ b/templates/utils.html
@@ -159,7 +159,7 @@
 			<li class="desktop_item"><a href="https://reddit.com{{ post.permalink }}" rel="nofollow">reddit</a></li>
 			<li class="mobile_item"><a href="https://reddit.com{{ post.permalink }}" rel="nofollow">reddit</a></li>
 		</ul>
-		<p>{{ post.upvote_ratio }}%<span id="upvoted"> Upvoted</span></p>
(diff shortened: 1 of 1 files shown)


Daniel-Valentine:
I like this. Only complaint is that the count doesn't show up in mobile view, which I think we want. If you address that, then I think this will be good to merge.

@spikecodes: thoughts?

spikecodes:
I like this idea, but I agree that it should also show on mobile.

One idea is displaying the comment count in line with the sorting algorithm selector:

image

What are your thoughts on this?

Daniel-Valentine:
@spikecodes: I prefer that approach.

sigaloid:
Agreed, that looks better. Feel free to overwrite my branch or create a new PR if you'd like!

Issue #444, opened 2022-03-03
libreddit is hard to use with screen readers and I have ideas for improvements.


crentava:
Hello.
I discovered libreddit recently and I like what it's trying to do. There is 1 problems I would like to address, however, that makes it inconvenient to use as a person who has to use a screen reader. This is easily fixable.

## lack of headings

The first problem I would like to talk about and is the main usability issue for us, there are no headings.
As a screen reader user, I use headings to navigate webpages because it is more convenient than using the arrow keys or the tab key to navigate around the page. Navigating with the arrow or tab keys is tedious and takes a lot of time to find what you are looking for. I will include a list of places where headings may be useful

* post titles on subreddits and the front and search pages, to make navigation between different posts easier.
* in the settings page to separate categories such as appearance, interface and all the other categories

There are also 2 other things I want to mention that will make navigation faster for us. One is the addition of links to skip to the main part of the page as well as the search field. The other thing would be putting the sort options into a combo box. This would make the page more compact and if anyone navigates by link, they will be able to reach what they want more quickly.

If you are worried about the change in appearance, You could include a checkbox in the settings page that says "enable accessibility mode" which would redirect to a template that has all of those changes.
thank you

spikecodes:
Hi,

Thank you for raising these issues; accessibility for screen readers isn't actively considered in most web development so I appreciate you sharing both your difficulties. and possible solutions.

1. I totally see what you mean about headings and will definitely convert some of the `<p>` tags like the post titles, subreddit titles, and user titles to `<h1>`s.
2. The settings page now has headings defined with the `<legend>` elements. Let me know if you think this is an incomplete solution.
3. I'm not sure how I'd implement a "skip to main content" button discretely and easily so hope that's not too big of an issue if not included.
4. Unfortunately, it wouldn't be possible to compress sort options into a combo box as that would break functionality for no-JS users.

Hope this answered some of your questions! I'll be patching 1) soon and am welcome to any other ideas.

crentava:
Thank you for taking my suggestions into consideration! The headings are the major thing to think about, so your ok :)

PR #662, opened 2022-12-14
fix a11y and HTML issues on settings page

rangermeier:
The settings page had some accessibility issues and contained HTML that didn't adhere to the spec. This PR fixes the following things

- Connect labels with corresponding form controls: the `for` attribute of the `<label>`s pointed to the ID of their parent `<div>`s. I removed the ID from the parent elements and added it to the corresponding `<select>` or `<input>`. Users can now click on the label to toggle the state of checkboxes and screenreaders can label the form elements.
- Use fieldsets to group form sections: #444 introduced `<legend>` elements as headers for the form's sections. The HTML spec allows `<legend>` only as direct child of a `<fieldset>`.
- Don't nest details/summary element into label: A `<label>` should not contain other interactive elements.

These changes don't change the visual appearance of the settings form.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -1118,22 +1118,16 @@ summary.comment_data {
 }
 
 .prefs {
-	display: flex;
-	flex-direction: column;
-	justify-content: space-between;
-	padding: 20px;
+	padding: 10px 20px 20px;
 	background: var(--post);
 	border-radius: 5px;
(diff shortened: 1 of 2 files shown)


Daniel-Valentine:
Approved. Thanks for your contribution. Sorry I missed this before I cut v0.25.2, but this will make it into the next release.

PR #663, opened 2022-12-16
Search - add support for raw reddit links

gmnsii:
This allows users to paste a Reddit (or old Reddit) link in the search bar and be redirected to the corresponding Libreddit page.

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -7,6 +7,8 @@ use crate::{
 };
 use askama::Template;
 use hyper::{Body, Request, Response};
+use once_cell::sync::Lazy;
+use regex::Regex;
 
 // STRUCTS
 struct SearchParams {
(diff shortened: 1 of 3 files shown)


sigaloid (approved):
Looks good to me, thanks for the contribution!

src/search.rs
```diff
@@ -50,7 +50,13 @@ struct SearchTemplate {
 pub async fn find(req: Request<Body>) -> Result<Response<Body>, String> {
 	let nsfw_results = if setting(&req, "show_nsfw") == "on" { "&include_over_18=on" } else { "" };
 	let path = format!("{}.json?{}{}&raw_json=1", req.uri().path(), req.uri().query().unwrap_or_default(), nsfw_results);
-	let query = param(&path, "q").unwrap_or_default();
+	let mut query = param(&path, "q").unwrap_or_default();
+
+	if query.starts_with("https://www.reddit.com/") {
```
Daniel-Valentine, line 55: This is a good start, but we need to support more than just `www.reddit.com` and `old.reddit.com`. Besides supporting `reddit.com` (no subdomain) we would need to support all of the other subdomains. `np.reddit.com` is an obvious one that comes to mind. I couldn't find a formal list of subdomains, so I think we should instead aim to determine of the query begins with `*.reddit.com` and strip the scheme and domain from that.

I think we should also support the `http` scheme as well if someone wants to put in a link URL that was generated before Reddit employed universal TLS.

In regex, what we'd look for is this prefix: `^https?://([^\./]+\.)*reddit.com/`

Daniel-Valentine (changes requested):

src/search.rs
Daniel-Valentine: If you want to avoid using `regex::Regex` unless absolutely necessary, you could write code that checks for the presence of the scheme, and if it is detected, then filter the string prefix:

```rust
if query.starts_with("https://") || if query.starts_with("http://")
```

Note the check on `https` comes first since nowadays `https` is a much more common scheme.

gmnsii: I went with a regex (see da6b716e7587987349e0643cb932cf8f5d76fb19) but if manual parsing is preferred just say so and I will change it.

Daniel-Valentine (approved):

Daniel-Valentine:
LGTM. Thanks for your contribution!

PR #664, opened 2022-12-17
Add config system to read from file 

sigaloid:
Currently, precedence is given to named environment variables over the config file - I believe this is the standard order of precedence but this can be changed. 

Next, we are going to implement a footer and API endpoint that returns some instance information.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,130 @@
+use once_cell::sync::Lazy;
+use std::{env::var, fs::read_to_string};
+
+// Waiting for https://github.com/rust-lang/rust/issues/74465 to land, so we
+// can reduce reliance on once_cell.
+//
+// This is the local static that is initialized at runtime (technically at
+// first request) and contains the instance settings.
+static CONFIG: Lazy<Config> = Lazy::new(Config::load);
+
+/// Stores the configuration parsed from the environment variables and the
+/// config file. `Config::Default()` contains None for each setting.
+#[derive(Default, serde::Deserialize)]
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -0,0 +1,130 @@
...
+	}
+}
+
+/// Retrieves setting from environment variable or config file.
+pub(crate) fn get_setting(name: &str) -> Option<String> {
+	get_setting_from_config(name, &CONFIG)
+}
+
+#[cfg(test)]
+use {sealed_test::prelude::*, std::fs::write};
```
Daniel-Valentine, line 100: I'm ok with this for now but I think we should have our test code live in a `mod test` rather than have them share the namespace with the rest of the crate.

But I don't care enough to request that you change this.

Daniel-Valentine (approved):
I reviewed this in its draft stage and declared this good to go in, so I'll approve it. Will leave it to @spikecodes to render a final decision.

src/config.rs
sigaloid: I agree, but I was following the example set in the `utils.rs` tests: https://github.com/libreddit/libreddit/blob/master/src/utils.rs#L832

I'll create an issue though - I agree that it should be moved out in the future.

Daniel-Valentine:
I've gone ahead and resolved the merge conflict; with that addressed, this can now go in, also per the agreement we've three had. I forgot to bring this in before cutting v0.26.0 (my bad), but we can definitely make this available in v0.26.1 or later.

PR #666, opened 2022-12-19
Bugfix: 'all posts are hidden because NSFW' when no posts where found

gmnsii:
Fix a bug where Libreddit would display a message saying 'All posts are hidden because they are NSFW, enable "Show NSFW posts"' when they were no posts.
Also add a specific message when no posts are found.


Daniel-Valentine:
Thanks for raising PR \#666. 😈

I was going to address this same issue as part of #656, but that might still be a few days out at the very least, so I think this is a good stopgap for now. Give me a chance to review, but I think this will be good to go in.

Daniel-Valentine (approved):

PR #667, opened 2022-12-26
Theme browser scrollbar

erdnaxe:
Hint current color-scheme to the browser.
This enables chromium-based browsers to change the scrollbar color according to the current theme.


sigaloid:
Thanks for the contribution! :)

PR #673, opened 2023-01-01
Link to `libreddit/libreddit` and open in new tab

spenserblack:
This sets the target of the "code" link to `_blank`, so that it will
open in a new tab in browsers. Because the GitHub page is a different
context from libreddit, and accessing the repository doesn't imply that
the user is finished browsing libreddit, this seemed reasonable. This
also changes the link from `spikecodes/libreddit` to
`libreddit/libreddit`.

******

I'm breaking my self-imposed rule of "one effective change per commit", but with a 1-line HTML diff
that didn't seem to matter :laughing:
Depending on how you prefer your diffs/history, I'm happy to change the history of this PR.


sigaloid:
Looks great, thanks for the contribution! I'm going to merge this - I don't think we'll have any objections by the other maintainers.

PR #674, opened 2023-01-01
Create devcontainer.json

spenserblack:
This helps contributors work on libreddit in a GitHub Codespace. Some explanation:
- More can be read about the image [here](https://github.com/devcontainers/images/tree/main/src/rust)
- The docker-in-docker feature allows usage of `docker` in the codespace
- When starting the app on port 8080, it will be labeled "libreddit" and the user will see a notification to open the page in a new tab
- Once the codespace is created, `cargo build` is run. This is somewhat inspired by the default Rust build steps of a gitpod instance. It's not that important, IMO, but the initial build can be kind of lengthy, so it can save a bit of time to have the build already running when you start developing.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,14 @@
+{
+    "name": "Rust",
+    "image": "mcr.microsoft.com/devcontainers/rust:0-1-bullseye",
+    "features": {
+        "ghcr.io/devcontainers/features/docker-in-docker:2": {}
+    },
+    "portsAttributes": {
+        "8080": {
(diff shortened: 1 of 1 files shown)


spenserblack:
Oh, also, I know that it's much more common for a `.json` file to have 2-space indentation, it seems like there's a slight preference 4-space indentation for this file. For example, the documented example usage for `devcontainers/features`. But, looking through your other files, it seems like tab indentation is heavily favored, so let me know if you'd like the indentation to be changed (an EditorConfig file might be helpful).

sigaloid:
Thanks for the contribution :partying_face:

Issue #675, opened 2023-01-01
🐛 Bug Report: 7-digit post IDs do not work.


ellieeet123:
## Describe the bug

All reddit posts have a base-36 ID to identify them. You can visit posts using this message, for example libreddit.spike.codes/<POST ID>. For 5 and 6 character IDs, this works exactly as expected. However, recently all of the 6 character IDs were used up, and newer posts have seven character IDs. On libreddit, attemping to view a seven character ID post via the ID only URL does not work. 
## Steps to reproduce the bug
 View this post: https://libreddit.spike.codes/100ryvp
The post can be viewed correctly here: https://libreddit.spike.codes/r/nextfuckinglevel/comments/100ryvp/fireworks_show_in_san_diego_where_a_computer/

## What's the expected behavior?
The post with the 7-character ID should be displayed.

spikecodes:
The issue appears to be on Reddit's end at the moment as they don't resolve 7-digit post IDs yet.

Example post: https://www.reddit.com/r/AskReddit/comments/101kh2j/boys_be_honest_what_makes_a_girl_instantly/

Doesn't get resolved: https://www.reddit.com/101kh2j

ellieeet123:
Oh, interesting. I didn't think to check that. I guess for now we will just have to wait

spikecodes:
The ID-matching code in Libreddit has been future-proofed to support 7-digit post IDs once Reddit adds redirect functionality. Closing this issue for now

PR #676, opened 2023-01-01
Fix for #675

ellieeet123:
/:id route now accepts 7 character post IDs.


sigaloid:
Wow, since this broke right at the new year's, I declare this to be Libreddit's Y2023 problem. **Thanks for the fix** - I'm merging this now as it's an important bugfix.

PR #679, opened 2023-01-02
[Awards] Add an option to display awards.

### LINKED ISSUE #442: Make "Awards visible" optional

The Awards are not essential to the discussion and take extra requests.

Optional setting to make awards visible on the interface with it being off by default. On both posts and comments

### Screenshot of requests on `r/popular`
Screen

annoying little critters

DiegoMagdaleno:
While this implementation is not very clean (code-wise), it does ensure that, the awards aren't even parsed if the option to display awards is disabled.

I am sure this implementation could be improved, currently, the biggest challenge is accessing the settings without a request.

My question is, wouldn't it be better to use a singleton pattern for the settings?

If we used a singleton pattern for the settings, we could have a way to access a same instance of the config anywhere in the code. We could update the singleton values, if and only if, the user made changes in the "settings" page. Per my understanding of the code, currently, this request to read the settings is done every time a post is parsed, I am sure we might be able to benefit from only requesting setting "parsing" when necessary.

Anyway, this fixes #442  👍🏼 

I am open for suggestions and ways to improve the code, as I don't think the implementation is clean enough, but it could benefit from the singleton pattern I mentioned above.

### DIFF
--- a/src/duplicates.rs
+++ b/src/duplicates.rs
@@ -3,7 +3,7 @@
 use crate::client::json;
 use crate::server::RequestExt;
 use crate::subreddit::{can_access_quarantine, quarantine};
-use crate::utils::{error, filter_posts, get_filters, parse_post, template, Post, Preferences};
+use crate::utils::{error, filter_posts, get_filters, parse_post, template, Post, Preferences, setting};
 
 use askama::Template;
 use hyper::{Body, Request, Response};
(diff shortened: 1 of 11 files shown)


DiegoMagdaleno:
I just realized someone already had something like this, however, I think this implementation is a little bit cleaner as it won't even parse the awards if the option to do so isn't selected.

DiegoMagdaleno:
Adding to this, I am willing to experiment with trying to implement a singleton pattern in order to reduce the number of times the settings get checked out :p

Daniel-Valentine:
Thanks for your contribution! We've decided to go with #678 because it offers an option to disable awards rather than making viewing awards opt-in.

PR #681, opened 2023-01-02
Make CLI args have higher precedence than env vars

spenserblack:
This simplifies the logic to build the listener by using more clap
features instead of manually accessing the PORT environment variable.
This also removes unnecessary `unwrap_or` calls that set defaults that
are already set by clap.

******

I stumbled on this while looking through the source code, and it looked like this part could have been simplified. I'm much more familiar with clap 2 than 3 or 4, so a review is appreciated!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2021"
 askama = { version = "0.11.1", default-features = false }
 async-recursion = "1.0.0"
 cached = "0.40.0"
-clap = { version = "4.0.24", default-features = false, features = ["std"] }
+clap = { version = "4.0.24", default-features = false, features = ["std", "env"] }
 regex = "1.7.0"
 serde = { version = "1.0.147", features = ["derive"] }
 cookie = "0.16.1"
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,7 +12,7 @@ mod user;
 mod utils;
 
 // Import Crates
-use clap::{Arg, Command};
+use clap::{Arg, ArgAction, Command};
 
 use futures_lite::FutureExt;
 use hyper::{header::HeaderValue, Body, Request, Response};
@@ -129,8 +129,10 @@ async fn main() {
 				.short('p')
 				.long("port")
 				.value_name("PORT")
+				.env("PORT")
 				.help("Port to listen on")
 				.default_value("8080")
+				.action(ArgAction::Set)
 				.num_args(1),
 		)
 		.arg(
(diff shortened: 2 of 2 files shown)


sigaloid:
Unfortunately I believe we have to check for the existence of the environment variable, as many users set the PORT variable with a Docker command and that's likely easier than manually setting the command line flags through Docker.

spenserblack:
I believe the `env` clap feature and `.env` will handle that for you. I did an informal, manual test by running `PORT=1234 cargo run -- --port 2345`, `PORT=1234 cargo run`, and `cargo run -- --port 2345`, and `cargo run`, and got expected results (`2345`, `1234`, `2345`, and `8080`). Though I didn't try it out with the docker image specifically.

sigaloid:
Ah, I see. This changes the precedence from environment variable-first to CLI-first if both arguments are provided - is there a way to retain the original behavior (environment variable is of highest precedence)?

spenserblack:
Oh, I see, I completely spaced out on the fact that env vars had higher priority :sweat_smile:

TBH I find env vars taking higher priority over CLI args to be unusual behavior. I think all tools I've used have done the reverse. I think that's because env vars are sometimes explicit, but CLI args are pretty much *always* explicit. For example, if I had `export PORT=1234` in my `.bashrc`, and I called `libreddit --port 2345`, I'd be a bit surprised that `1234` was used. But if env vars taking higher priority is the behavior your users expect and prefer, I definitely don't want to change that! :slightly_smiling_face:

It will shrink the significance of this PR, but I suppose it could be changed to just remove the `unwrap_or`s. It's not a big deal, but they're a bit redundant when those values are already set as defaults on the `clap::Arg`s.

sigaloid:
That's a good point. @Daniel-Valentine @spikecodes Thoughts on changing preference to command line? It typically goes CLI > env > config file so this makes sense to change IMO. Though there is no equivalent config file option for the port variable

spenserblack:
One thing you might want to look into the future is config-rs. It abstracts reading configuration from environment variables and a config file for you, and I think the maintainer is considering adding support for clap, too.

sigaloid:
We attempted to use that crate, but it required renaming some of the config settings, and we didn't want to cause that kind of change for little benefit (we really only need one config format).

spikecodes:
Thanks for this PR @spenserblack!

> That's a good point. @Daniel-Valentine @spikecodes Thoughts on changing preference to command line? It typically goes CLI > env > config file so this makes sense to change IMO. Though there is no equivalent config file option for the port variable

I'm fine with that order of preference. It's good that CLI > env in case `PORT` is already defined for another self-hosted application.

spikecodes (approved):

Daniel-Valentine:
Sorry I didn't get to this earlier. All of this looks good to me.

Issue #347, opened 2021-11-28
Add command line argument for instance-wide banner


spikecodes:
## Is your feature request related to a problem? Please describe.
N/A

## Describe the solution you'd like
Allow hosters of Libreddit to display a banner where users can donate to the hosters or learn something about the service. Example:
```bash
libreddit -b "This instance may be slow due to many users browsing it. We recommend using another instance."
```

There should be an ability to permanently hide them for each instance though.

## Describe alternatives you've considered
N/A

## Additional context
Invidious has a similar feature for hosters to list messages at the top of each page.

PR #685, opened 2023-01-03
Implement instance info endpoint (JSON, YAML, TXT)

### LINKED ISSUE #624: Detail instance info in a REST endpoint or existing page metadata
Add data in the `<head>` block that details information about the Libreddit instance. Example, using [`<meta name="application-name">`](https://html.spec.whatwg.org/multipage/semantics.html#meta-application-name):

```jinja2
<meta name="application-name"
    content="Libreddit"
    data-version="{{ env!("CARGO_PKG_VERSION") }}"
    data-hls-enabled="false"
    data-default-theme="dark"
    data-sfw="true"
    data-sfw-only="false"
    data-content-encoding="gzip"
/>
```

This is not the only way we could add info to `<head>`. Per [this Stack Exchange post](https://stackoverflow.com/a/19298638), we have other options, including defining a `<script type="text/plain">` block with content of our own encoding (e.g. CSV). That might even be preferable to a `<meta>` tag.

The purpose for this is so that instance scrapers, like [libreddit/libreddit-instances](https://github.com/libreddit/libreddit-instances), can get information about them without having to rely on hard-coded data, e.g. the CSV in the libreddit-instances repo. This would help address libredirect/libredirect#477.

sigaloid:
## https://685.fly.dev/

Fixes #624 

I added a feature to the `time` crate to allow for local timezone loading. I also added `serde_yaml` so we could return it in YAML form, as well as JSON and regular text; as well as `build_html` in order to more easily construct an HTML table. These are quite small dependencies. Here's what they look like: 

## HTML (default with /instance-info)
image

## JSON
image

## YAML 
image

## TXT
image

The `build.rs` is required and standard practice to set a variable - AFAIK it's the easiest way to embed the latest commit hash. 

Within `instance_info.rs`, I made some choices - the `InstanceInfo` struct will be held in a Lazy. Currently nothing in the struct should change so it's okay to cache it. However, from #624: 

> * Maybe some kind of latency statistics? Average latency to reddit, average response latency, etc.
>  * Number of responses served

If we add either of these, it will change on the fly, so we need to revisit the use of a `Lazy<InstanceInfo>`.

Within the `instance_info` function, I choose `html` as the default.

In `main.rs`, I forced evaluation of the two relevant statics (config and instance info) to ensure deploy timestamp is accurate, and to avoid paying the penalty of parsing the config at the first request.

In conclusion, this PR will allow us to implement a system where we can easily check if an instance is actually online and lets us measure versions across the instance ecosystem. If we add metrics like average Reddit latency or number of responses served, we can more accurately predict the performance of certain servers and bias a redirect server (going to work on this in the coming days :wink:) towards those.

I also added a link to the instance info page into `base.html`: 
image

The groundwork is also laid for a banner #347  - it's been included in the parsing code and is shown on the instance page. We need to discuss whether we want to display it on regular pages (footer? header? where? etc) or just on the info page.

I also added a "LIBREDDIT_DEFAULT_HIDE_AWARDS" option for a default - in case a server operator wants to hide it by default (goes with #678).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,6 @@
+use std::process::Command;
+fn main() {
+	let output = Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap();
+	let git_hash = String::from_utf8(output.stdout).unwrap();
+	println!("cargo:rustc-env=GIT_HASH={git_hash}");
+}
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,20 @@
-use std::process::Command;
+use std::{
+	os::unix::process::ExitStatusExt,
+	process::{Command, ExitStatus, Output},
+};
 fn main() {
-	let output = String::from_utf8(Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap().stdout).unwrap_or_default();
+	let output = String::from_utf8(
+		Command::new("git")
+			.args(["rev-parse", "HEAD"])
+			.output()
+			.unwrap_or(Output {
+				stdout: vec![],
+				stderr: vec![],
+				status: ExitStatus::from_raw(0),
+			})
+			.stdout,
+	)
+	.unwrap_or_default();
 	let git_hash = if output == String::default() { "dev".into() } else { output };
 	println!("cargo:rustc-env=GIT_HASH={git_hash}");
 }
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,6 @@
 use std::process::Command;
 fn main() {
-	let output = Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap();
-	let git_hash = String::from_utf8(output.stdout).unwrap();
+	let output = String::from_utf8(Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap().stdout).unwrap_or_default();
+	let git_hash = if output == String::default() { "dev".into() } else { output };
 	println!("cargo:rustc-env=GIT_HASH={git_hash}");
 }
--- a/src/instance_info.rs
+++ b/src/instance_info.rs
@@ -1,7 +1,7 @@
 use crate::{
 	config::{Config, CONFIG},
 	server::RequestExt,
-	utils::Preferences,
+	utils::{ErrorTemplate, Preferences},
 };
 use askama::Template;
 use build_html::{Container, Html, HtmlContainer, Table};
@@ -21,10 +21,20 @@ pub async fn instance_info(req: Request<Body>) -> Result<Response<Body>, String>
 	// simply become the last option, an HTML page.
 	let extension = req.param("extension").unwrap_or(String::new());
 	let response = match extension.as_str() {
-		"yaml" => info_yaml(),
+		"yaml" | "yml" => info_yaml(),
 		"txt" => info_txt(),
 		"json" => info_json(),
-		"html" | _ => info_html(req),
+		"html" | "" => info_html(req),
+		_ => {
+			let error = ErrorTemplate {
+				msg: "Error: Invalid info extension".into(),
+				prefs: Preferences::new(&req),
+				url: req.uri().to_string(),
+			}
+			.render()
+			.unwrap();
+			Response::builder().status(404).header("content-type", "text/html; charset=utf-8").body(error.into())
+		}
 	};
 	response.map_err(|err| format!("{err}"))
 }
@@ -83,31 +93,41 @@ impl InstanceInfo {
 	}
 	fn to_table(&self) -> String {
 		let mut container = Container::default();
-		let convert = |o: &Option<String>| -> String { o.clone().unwrap_or("Unset".to_owned()) };
-		container.add_header(3, "Instance banner");
+		let convert = |o: &Option<String>| -> String { o.clone().unwrap_or("<span class=\"unset\"><i>Unset</i></span>".to_owned()) };
+		if let Some(banner) = &self.config.banner {
+			container.add_header(3, "Instance banner");
+			container.add_raw("<br />");
+			container.add_paragraph(banner);
+			container.add_raw("<br />");
+		}
+		container.add_table(
+			Table::from([
+				["Crate version", &self.crate_version],
+				["Git commit", &self.git_commit],
+				["Deploy date", &self.deploy_date],
+				["Deploy timestamp", &self.deploy_unix_ts.to_string()],
+				["Compile mode", &self.compile_mode],
+			])
+			.with_header_row(["Settings"]),
+		);
 		container.add_raw("<br />");
-		container.add_paragraph(convert(&self.config.banner));
-		container.add_raw("<br />");
-		container.add_table(Table::from([
-			["Crate version", &self.crate_version],
-			["Git commit", &self.git_commit],
-			["Deploy date", &self.deploy_date],
-			["Deploy timestamp", &self.deploy_unix_ts.to_string()],
-			["Compile mode", &self.compile_mode],
-			["<b>Settings</b>", "<b>Settings</b>"],
-			["SFW only", &convert(&self.config.sfw_only)],
-			["Hide awards", &convert(&self.config.default_hide_awards)],
-			["Default theme", &convert(&self.config.default_theme)],
-			["Default front page", &convert(&self.config.default_front_page)],
-			["Default layout", &convert(&self.config.default_layout)],
-			["Default wide", &convert(&self.config.default_wide)],
-			["Default comment sort", &convert(&self.config.default_comment_sort)],
-			["Default post sort", &convert(&self.config.default_post_sort)],
-			["Default show NSFW", &convert(&self.config.default_show_nsfw)],
-			["Default blur NSFW", &convert(&self.config.default_blur_nsfw)],
-			["Default use HLS", &convert(&self.config.default_use_hls)],
-			["Default hide HLS notification", &convert(&self.config.default_hide_hls_notification)],
-		]));
+		container.add_table(
+			Table::from([
+				["SFW only", &convert(&self.config.sfw_only)],
+				["Hide awards", &convert(&self.config.default_hide_awards)],
+				["Default theme", &convert(&self.config.default_theme)],
+				["Default front page", &convert(&self.config.default_front_page)],
+				["Default layout", &convert(&self.config.default_layout)],
+				["Default wide", &convert(&self.config.default_wide)],
+				["Default comment sort", &convert(&self.config.default_comment_sort)],
+				["Default post sort", &convert(&self.config.default_post_sort)],
+				["Default show NSFW", &convert(&self.config.default_show_nsfw)],
+				["Default blur NSFW", &convert(&self.config.default_blur_nsfw)],
+				["Default use HLS", &convert(&self.config.default_use_hls)],
+				["Default hide HLS notification", &convert(&self.config.default_hide_hls_notification)],
+			])
+			.with_header_row(["Default preferences"]),
+		);
 		container.to_html_string()
 	}
 	fn to_string(&self, string_type: StringType) -> String {
--- a/src/instance_info.rs
+++ b/src/instance_info.rs
@@ -1,0 +1,108 @@
+use hyper::{http::Error, Body, Request, Response};
+use once_cell::sync::Lazy;
+use serde::{Deserialize, Serialize};
+use time::OffsetDateTime;
+
+use crate::{
+	config::{Config, CONFIG},
+	server::RequestExt,
+};
+
+// This is the local static that is intialized at runtime (technically at
+// the first request *to the instance-info endpoint) and contains the data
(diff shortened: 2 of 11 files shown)


build.rs
```diff
@@ -0,0 +1,6 @@
+use std::process::Command;
+fn main() {
+	let output = Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap();
```
NickCao, line 3: For those building from release tarballs, git information is not available. Should either fallback to a fixed value like `dev` gracefully, or allow specifying a version via an environment variable.

sigaloid: Thank you for the review!

templates/base.html
```diff
@@ -66,6 +66,11 @@
 		</main>
 		{% endblock %}
 		{% block footer %}
+			<footer>
+				<div>
+					<a href="/instance-info">View instance information</a>
+				</div>
+			</footer>
 			{% if crate::utils::sfw_only() %}
 			<footer><div id="sfw-only">This instance of Libreddit is SFW-only.</div></footer>
```
Daniel-Valentine, line 75: We won't need the SFW-only disclosure in the footer anymore with the instance info page.

src/main.rs
```diff
@@ -283,6 +292,10 @@ async fn main() {
 	// Handle about pages
 	app.at("/about").get(|req| error(req, "About pages aren't added yet".to_string()).boxed());
 
+	// Instance info page
+	app.at("/instance-info").get(|r| instance_info::instance_info(r).boxed());
```
Daniel-Valentine, line 296:
```suggestion
	app.at("/info").get(|r| instance_info::instance_info(r).boxed());
```

I prefer `/info` since it's concise.

src/main.rs
```diff
@@ -283,6 +292,10 @@ async fn main() {
...
+	app.at("/instance-info.:extension").get(|r| instance_info::instance_info(r).boxed());
```
Daniel-Valentine, line 297: Consider handling invalid extensions. I just tried `/instance-info.jsn` by accident and got the HTML page, when we really should be returning a 404.

src/instance_info.rs
```diff
@@ -0,0 +1,171 @@
...
+		container.add_table(Table::from([
+			["Crate version", &self.crate_version],
+			["Git commit", &self.git_commit],
+			["Deploy date", &self.deploy_date],
+			["Deploy timestamp", &self.deploy_unix_ts.to_string()],
+			["Compile mode", &self.compile_mode],
+			["<b>Settings</b>", "<b>Settings</b>"],
+			["SFW only", &convert(&self.config.sfw_only)],
+			["Hide awards", &convert(&self.config.default_hide_awards)],
+			["Default theme", &convert(&self.config.default_theme)],
```
Daniel-Valentine, line 100: I think we should have a separate table for the instance's default user settings. This way, we don't have to prepend each of these setting names with "Default".

src/instance_info.rs
```diff
@@ -0,0 +1,171 @@
...
+			deploy_unix_ts: OffsetDateTime::now_local().unwrap_or_else(|_| OffsetDateTime::now_utc()).unix_timestamp(),
+			config: CONFIG.clone(),
+		}
+	}
+	fn to_table(&self) -> String {
+		let mut container = Container::default();
+		let convert = |o: &Option<String>| -> String { o.clone().unwrap_or("Unset".to_owned()) };
+		container.add_header(3, "Instance banner");
+		container.add_raw("<br />");
+		container.add_paragraph(convert(&self.config.banner));
```
Daniel-Valentine, line 89:
```suggestion
		if let Some(banner) = &self.config.banner {
				container.add_paragraph(convert(banner));
		}
```

This way, "Unset" won't be in the banner atop the table if it isn't actually set. This is currently the case on your fly.dev demo: https://685.fly.dev/instance-info

src/instance_info.rs
```diff
@@ -0,0 +1,171 @@
...
+			compile_mode: "Debug".into(),
+			#[cfg(not(debug_assertions))]
+			compile_mode: "Release".into(),
+			deploy_unix_ts: OffsetDateTime::now_local().unwrap_or_else(|_| OffsetDateTime::now_utc()).unix_timestamp(),
+			config: CONFIG.clone(),
+		}
+	}
+	fn to_table(&self) -> String {
+		let mut container = Container::default();
+		let convert = |o: &Option<String>| -> String { o.clone().unwrap_or("Unset".to_owned()) };
```
Daniel-Valentine, line 86:
```suggestion
		let convert = |o: &Option<String>| -> String { o.clone().unwrap_or("<i>(unset)</i>".to_owned()) };
```

To make it clearer the value is not set. I'd even recommend `<span>` and a CSS class to change the color of the text to something lighter to make it clear that the parameter is unset.

src/instance_info.rs
```diff
@@ -0,0 +1,171 @@
...
+// retrieved from the instance-info endpoint.
+pub(crate) static INSTANCE_INFO: Lazy<InstanceInfo> = Lazy::new(InstanceInfo::new);
+
+/// Handles instance info endpoint
+pub async fn instance_info(req: Request<Body>) -> Result<Response<Body>, String> {
+	// This will retrieve the extension given, or create a new string - which will
+	// simply become the last option, an HTML page.
+	let extension = req.param("extension").unwrap_or(String::new());
+	let response = match extension.as_str() {
+		"yaml" => info_yaml(),
```
Daniel-Valentine, line 24:
```suggestion
		"yaml" | "yml" => info_yaml(),
```

src/instance_info.rs
```diff
@@ -0,0 +1,171 @@
...
+		"txt" => info_txt(),
+		"json" => info_json(),
+		"html" | _ => info_html(req),
```
Daniel-Valentine, line 27:
```suggestion
		"html" => info_html(req),
```

I think we should have a `_` arm that returns our default 404.

Daniel-Valentine (changes requested):

templates/base.html
```diff
@@ -66,6 +66,11 @@
 		</main>
 		{% endblock %}
 		{% block footer %}
+			<footer>
+				<div>
+					<a href="/instance-info">View instance information</a>
```
Daniel-Valentine, line 71: This isn't bad, but is there any chance we can make this a little prettier? 🙂

Daniel-Valentine (approved):

src/instance_info.rs
Daniel-Valentine: blah

Daniel-Valentine (changes requested):

Daniel-Valentine (dismissed):
(leaving a comment to convince GitHub that I'm requesting changes, which I am)

templates/base.html
sigaloid: I tried to spruce it up - nothing too fancy but now it's got some shadows and reacts to the cursor. It'll be up at https://685.fly.dev in a few minutes when it finishes compiling, if you'd like to take a look.

spikecodes:
I've removed the hard-coded colors from the `.info-button` in favor of the style-agnostic `--text` color variable so it'll show in dark themes as well. I've also aligned the footer to the far right so that the info link doesn't conflict (in terms of design hierarchy) with the Next/Previous buttons that are center-aligned. I also removed the border and box shadow because for non-interactive links (like Settings or Info), I think an underline makes more sense and helps differentiate from the buttons.

Otherwise looks great to me! I think Daniel had a couple more comments but I'm looking forward to this being merged.

src/instance_info.rs
```diff
@@ -0,0 +1,191 @@
...
+				["Deploy date", &self.deploy_date],
+				["Deploy timestamp", &self.deploy_unix_ts.to_string()],
+				["Compile mode", &self.compile_mode],
+			])
+			.with_header_row(["Settings"]),
+		);
+		container.add_raw("<br />");
+		container.add_table(
+			Table::from([
+				["SFW only", &convert(&self.config.sfw_only)],
```
Daniel-Valentine, line 116: The SFW-only setting needs to go in the table above since it's an instance setting and not a user-configurable one.

src/instance_info.rs
```diff
@@ -0,0 +1,191 @@
...
+				["Hide awards", &convert(&self.config.default_hide_awards)],
+				["Default theme", &convert(&self.config.default_theme)],
```
Daniel-Valentine, line 118: Get rid `Default` in each of these rows.

templates/base.html
```diff
@@ -66,9 +66,11 @@
 		</main>
 		{% endblock %}
 		{% block footer %}
-			{% if crate::utils::sfw_only() %}
-			<footer><div id="sfw-only">This instance of Libreddit is SFW-only.</div></footer>
-			{% endif %}
+			<footer>
+				<div class="info-button">
+					<a href="/info">View instance information</a>
```
Daniel-Valentine, line 71: At least Firefox 108.0.2, this does not appear centered. I assume you want this to be centered?

src/instance_info.rs
```diff
@@ -0,0 +1,191 @@
...
+			.unwrap();
+			Response::builder().status(404).header("content-type", "text/html; charset=utf-8").body(error.into())
+		}
+	};
+	response.map_err(|err| format!("{err}"))
+}
+
+fn info_json() -> Result<Response<Body>, Error> {
+	let body = serde_json::to_string(&*INSTANCE_INFO).unwrap_or("Error serializing JSON.".into());
+	Response::builder().status(200).header("content-type", "application/json").body(body.into())
```
Daniel-Valentine, line 44: In the rare instance we can't serialize the JSON, this will result in an HTTP 200 response with this error message when it really should be 5xx.

src/instance_info.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+fn info_yaml() -> Result<Response<Body>, Error> {
+	let body = serde_yaml::to_string(&*INSTANCE_INFO).unwrap_or("Error serializing YAML.".into());
```
Daniel-Valentine, line 48: See my comment above regarding JSON serialization and the incorrect HTTP status code.

src/instance_info.rs
```diff
@@ -0,0 +1,191 @@
...
+		}
+		container.add_table(
+			Table::from([
+				["Crate version", &self.crate_version],
+				["Git commit", &self.git_commit],
+				["Deploy date", &self.deploy_date],
+				["Deploy timestamp", &self.deploy_unix_ts.to_string()],
+				["Compile mode", &self.compile_mode],
+			])
+			.with_header_row(["Settings"]),
```
Daniel-Valentine, line 111: At least on Firefox 108.0.2, the table header spans only the first column and not the entire row. This might be a limitation/bug of the crate and not a fault on your part.

image

src/instance_info.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+fn info_json() -> Result<Response<Body>, Error> {
+	let body = serde_json::to_string(&*INSTANCE_INFO).unwrap_or("Error serializing JSON.".into());
+	Response::builder().status(200).header("content-type", "application/json").body(body.into())
+}
+
+fn info_yaml() -> Result<Response<Body>, Error> {
+	let body = serde_yaml::to_string(&*INSTANCE_INFO).unwrap_or("Error serializing YAML.".into());
+	// https://github.com/ietf-wg-httpapi/mediatypes/blob/main/draft-ietf-httpapi-yaml-mediatypes.md
```
Daniel-Valentine, line 49:
```suggestion
	// We can use `application/yaml` as media type, though there is no guarantee
	// that browsers will honor it. But we'll do it anyway. See:
	// https://github.com/ietf-wg-httpapi/mediatypes/blob/main/draft-ietf-httpapi-yaml-mediatypes.md#media-type-applicationyaml-application-yaml

Daniel-Valentine (changes requested):

templates/base.html
sigaloid: @spikecodes changed this in abd24822 - I'm not sure whether this was intended. It was initially centered but maybe end-justified makes it less prominent on purpose. I'm not opposed to it either way, so it's up to you and Spike.

src/instance_info.rs
sigaloid: Yep - this is a limitation I can't directly change but can be patched around with a replace call.

templates/base.html
spikecodes: Yeah, the right-alignment was intentional. My thinking is that having a centered info button right below the centered Next/Previous buttons (in page feeds) is too much of a conflict in design hierarchy.

sigaloid: Gotcha, closing as resolved

Daniel-Valentine: I'm not sure I like this. To me, it looks like a CSS/HTML mistake rather than something consistent with Libreddit's design rhetoric. I'll see if I can come up with something that addresses @spikecodes' concern.

This issue otherwise, this is ready to go in.

Daniel-Valentine (approved):

Issue #689, opened 2023-01-04
LIBREDDIT_DEFAULT_SUBSCRIPTIONS setting is no longer read


lleddewk:
Since upgrading to v0.27.0, the $LIBREDDIT_DEFAULT_SUBSCRIPTIONS setting is not honoured as it was in previous versions. The result is that I see the a feed of random posts rather than the feeds that I am interested in. I think this was caused by the changes introduced by pull request #664.

Thanks
Keith

sigaloid:
This configuration was undocumented so I didn't know to include it in my PR. I'm going to fix this right after #685 goes in to avoid a conflict.

lleddewk:
Yeah I  realised that after posting so not surprising you didn't pick it up.  Grateful for the fix in due course.

Thanks
Keith

r7l:
Thanks allot for your work. Sadly it seems as if it didn't make it into the 0.27.1 release. Would be great to see it fixed at some point. For the time being i'm forced to stick with 0.26 as i don't want to loose this setting.

sigaloid:
Apologies, I was hoping we could get #685 in before 0.27.1 - I'm going to add it as soon as it goes in so we can get it for the next release which should be soon as we've been working through the PR. It's a sizable change so we want to make sure it is done right.

lleddewk:
Appreciate the efforts. As a workaround, you can save a bookmark to https://your.libreddit.site/r/sub1+sub2+sub3/.

Keith

r7l:
Great to see that blocking PR made it into the new release.

sigaloid:
Hi, apologies for the stagnation. I was planning on having this go right after 685 but we've cut a new release. 

I'm working on this right now!

PR #706, opened 2023-01-30
Fix default subscriptions, add option for pushshift frontend

### LINKED ISSUE #682: 💡 Feature request: Add option for Pushshift API site
Last year it was requested to switch from reveddit.com to unddit.com as it was considered slow. While that was the case back then, it's the opposite now. Unddit isn't working well recently. Since both of them are interchangeable by just using the other domain it would be great to have an option like `LIBREDDIT_PUSHSHIFT_FRONTEND` where you can set the domain.

sigaloid:
Fixes #682, #689

I combined these two as they're pretty similar - both config related and clearly defined. 

As per #682 perhaps we can consider changing the default to reveddit - is it faster? does it work better? etc.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -52,8 +52,14 @@ pub struct Config {
 	#[serde(rename = "LIBREDDIT_DEFAULT_HIDE_AWARDS")]
 	pub(crate) default_hide_awards: Option<String>,
 
+	#[serde(rename = "LIBREDDIT_DEFAULT_SUBSCRIPTIONS")]
+	pub(crate) default_subscriptions: Option<String>,
+
 	#[serde(rename = "LIBREDDIT_BANNER")]
 	pub(crate) banner: Option<String>,
+
+	#[serde(rename = "LIBREDDIT_PUSHSHIFT_FRONTEND")]
+	pub(crate) pushshift: String,
 }
 
 impl Config {
@@ -68,6 +74,11 @@ impl Config {
 		// environment variables with "LIBREDDIT", then check the config, then if
 		// both are `None`, return a `None` via the `map_or_else` function
 		let parse = |key: &str| -> Option<String> { var(key).ok().map_or_else(|| get_setting_from_config(key, &config), Some) };
+
+		// This serves as the frontend for the Pushshift API - on removed comments, this URL will
+		// be the base of a link, to display removed content (on another site).
+		let default_pushshift_frontend = String::from("www.unddit.com");
+
 		Self {
 			sfw_only: parse("LIBREDDIT_SFW_ONLY"),
 			default_theme: parse("LIBREDDIT_DEFAULT_THEME"),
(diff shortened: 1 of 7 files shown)


sigaloid:
Scratching my head here... :thinking: going to take a look at this later today

src/config.rs
```diff
@@ -52,8 +52,14 @@ pub struct Config {
...
 	pub(crate) default_hide_awards: Option<String>,
 
+	#[serde(rename = "LIBREDDIT_DEFAULT_SUBSCRIPTIONS")]
+	pub(crate) default_subscriptions: Option<String>,
+
 	#[serde(rename = "LIBREDDIT_BANNER")]
 	pub(crate) banner: Option<String>,
+
+	#[serde(rename = "LIBREDDIT_PUSHSHIFT_FRONTEND")]
+	pub(crate) pushshift: String,
```
r7l, line 62: @sigaloid Should this be the cause for the failed test? Looks like as if this should be `Option<String>`?

sigaloid: It needs to be String, as there must be a frontend specified. In any case I don't think that could really make it fail, because it's not affecting the existing test cases... No new tests were added. :/

r7l: Ok, i see. I was just trying to have a look at the failed test case but i have to admit that i never worked with Rust before. Just saw that the `get_setting_from_config` has `-> Option<String>` and i would expect that function to always return `Option<String>` which is no longer the case due to `LIBREDDIT_PUSHSHIFT_FRONTEND` being different.

If i may ask a question: How is `LIBREDDIT_BANNER` different from `LIBREDDIT_PUSHSHIFT_FRONTEND`? It is even [defined to be string in the README](https://github.com/libreddit/libreddit#instance-settings).

sigaloid: Banner is optional - if there's none set, it won't be displayed. It's listed as string, because not everyone reading the readme would know what `Option<String>` means

Pushshift frontend needs to be defined - it's essentially defined by default with a String rather than an Option<String> to let it fit better in the template

r7l: @sigaloid Looks like i can't help here much even if i would love to. But beside Libreddit, i've never had a look at Rust code. Would love to have this in Libreddit but it seems stuck for now. Anyways, thanks allot for explaining and your work on this PR and Libreddit.

r7l: @sigaloid Would it be an option to split this PR and try one for the default subscriptions only? It would be great to have this feature back in.

Tokarak:
Found the error using this debug code!
```rust
let config_to_write = r#"LIBREDDIT_DEFAULT_COMMENT_SORT = "best""#;
	write("libreddit.toml", config_to_write).unwrap();
	dbg!(toml::from_str::<Config>(&read_to_string("libreddit.toml").unwrap()));
```
Because `LIBREDDIT_PUSHSHIFT_FRONTEND` is non-optional, it is required during deserialisation. 
It's best to follow standard project practice: use an optional type (the user may or may not set this config), and use `unwrap_or` to fallback into a default value.

It may be also be a good idea to add a test for deserialising from an empty string to avoid this error in the future.

Issue #697, opened 2023-01-17
✨ Feature parity: Allow viewing of gated subreddits


rioting1312:
## How does this feature work on Reddit?
Usually, you agree to view a gated subreddit, and then you can view it. On libreddit, it just says, "gated", and you can't proceed further to view the content.

## Describe how this could be implemented into Libreddit
Allow the server hosting a libreddit instance to access the gated content, and then serve it to the viewers accessing the instance after displaying the gate's specific content, if the viewer agrees to the gate.

## Describe alternatives you've considered
Well, there currently isn't one, for people who want to use libreddit to view gated subreddits...

sigaloid:
Hi, I'm not sure what you mean by "gated". Can you provide an example of a gated subreddit?

rioting1312:
all the drug-related subreddits were recently "gated". This functionally broke libreddit from proxying these pages. You can see why users might want to use libreddit for these type of pages; they are more likely to value their privacy, and be suspicious of reddit's intention of surveilling them.

https://www.reddit.com/r/drugs


Jan 17, 2023, 06:17 by ***@***.***:

sigaloid:
Ah, okay. Thanks for the report. I'm going to investigate later - I don't think it's going to require a special bypass, like an account cookie from the server operator. On old.reddit.com it's available just by pressing view (same as NSFW) but I guess the JSON route requires some parameter or header not already set.

Daniel-Valentine:
The `gated` message, laconic as it is, comes directly from Reddit via their JSON API. The same holds true for `banned` (and other strings I can't remember). What we can do for now is change this message to something more explicit, like

> Reddit has gated this content and made it unavailable for viewing.

Same thing for `banned`

> This post cannot be viewed because it is on a subreddit that is banned.

(or perhaps just use the same message that appears when trying to access a banned subreddit: "r/[sub] has been banned from Reddit")

Daniel-Valentine:
We do need to be mindful of the [Reddit API TOS](https://archive.ph/qYYrT), specifically 3(b)(iii):

> 3.  Fees; Restrictions on Use. 
>     b.  Restrictions.  You must not, and must not allow those acting on your behalf to:
[...]

PR #722, opened 2023-02-17
Make gated subreddits accessible by treating them as quarantined

domve:
Addresses #697, this would be at least a temporary fix to make gated communities such as r/drugs accessible, as right now there's no way to get past the `gated` messages.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -140,7 +140,7 @@ fn request(method: &'static Method, path: String, redirect: bool, quarantine: bo
 		.header("Accept-Encoding", if method == Method::GET { "gzip" } else { "identity" })
 		.header("Accept-Language", "en-US,en;q=0.5")
 		.header("Connection", "keep-alive")
-		.header("Cookie", if quarantine { "_options=%7B%22pref_quarantine_optin%22%3A%20true%7D" } else { "" })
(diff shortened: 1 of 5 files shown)


vlfldr:
This is an important change. Large swaths of reddit have been inaccessible via libreddit for roughly a month. Thanks for authoring it @domve and thanks in advance to whomever reviews it!

sigaloid:
Thank you for this PR!

I think it might be a good idea to separate "quarantined" from "gated" - maybe pass the string into `quarantine` like so:

```rust
pub fn quarantine(req: Request<Body>, sub: String, restriction: String) -> Result<Response<Body>, String> {
	let wall = WallTemplate {
		title: format!("r/{} is {restriction}", sub),
```
The restriction is "quarantined" or "gated", so this works okay.

Also, there's a few other checks in the code for "quarantined" - perhaps those should be checked as well? (search `if msg == "quarantined"` in the code)

domve:
Thanks for the heads up, added checks for the rest of those. 

I've also implemented the restriction thing in `quarantine`, I think it would be worth displaying the warning messages too but it might be neater for me to open a separate PR for it after this gets merged.

spikecodes (approved):
Looks good to me! Thanks for this PR

Daniel-Valentine:
Looks good to me. Thanks for your contribution!

Issue #718, opened 2023-02-12
💡 Feature request: Allow overriding the instance webroot 


spikecodes:
## Is your feature request related to a problem? Please describe.
N/A

## Describe the feature you would like to be implemented
Allow hosters to pass a setting when launching Libreddit that overrides the root used for routing instead of the default `/`. This could allow hosting Libreddit on a subpath of a website. For instance, with a  webroot of `/libreddit` set, one could access the Rust subreddit at `<INSTANCE DOMAIN>/libreddit/r/rust`.

This could be implemented by prepending each route in [`main.rs`](https://github.com/libreddit/libreddit/blob/master/src/main.rs) with the defined webroot.

## Additional context / screenshot
This is possible when hosting Nextcloud with the `overwritewebroot` setting. I believe SearX and Vaultwarden support something similar.

Suggested by 0x4e4448 in our Matrix room.

Issue #702, opened 2023-01-25
💡 Feature request: Support listening on a UNIX socket


r3g-5z:
## Is your feature request related to a problem? Please describe.
A great use-case for UNIX sockets are for finer-grained access controls. For example, I only want my reverse proxy to see and communicate with libreddit and nothing else instead of listening on localhost which inherently allows other things to see and access libreddit in some way. This allows further isolation and sandboxing of libreddit which goes very well with a highly confined and restrictive systemd unit. This would even allow me to drop `AF_INET` and `AF_INET6` as allowed address families and keep it purely `AF_UNIX`.

UNIX sockets are also technically faster (especially throughput) than going through TCP/IP stack for loopback networking, but this is more relevant in performance-critical things like databases and the real bottleneck would be some place else like the reverse proxy or application itself rather than the transport method for something like libreddit.

## Describe the feature you would like to be implemented
I'd like libreddit to allow listening and binding to a UNIX socket to an arbitrary path (e.g. `/run/libreddit/libreddit.sock`) instead of listening over TCP on loopback. This could possibly be done by checking if the `ADDRESS` variable is a filesystem path, either to a directory (`/run/libreddit`) where it will create a `libreddit.sock` file, or a full path including file name (`/run/libreddit/libreddit.sock`).

The user will be up to themselves to configure permissions and access controls, but it is trivial to do.

## Describe alternatives you've considered
I considered using [systemd's PrivateNetwork](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#PrivateNetwork=) sandboxing and joining the reverse proxy's namespace (Caddy in my case) so libreddit can't see anything else and nothing else can see libreddit unless it's in libreddit's namespace along with restricting all networking to just localhost, but libreddit seems to make its own outbound traffic presumably to Reddit so I can't do that. PrivateNetwork also inherently restricts outbound networking since it only creates a loopback interface.

## Additional context / screenshot
https://quantum5.ca/2021/06/05/sharing-unix-sockets-between-multiple-users/
https://www.baeldung.com/linux/communicate-with-unix-sockets

I have searched the issue tracker and code and was not able to find any indicators that this feature is implemented or someone has previously requested this.

Daniel-Valentine:
This is an excellent idea and I'd be happy to implement this.

PR #727, opened 2023-02-19
Reuse hyper client.

wsy2220:
Making a new connection on every request is very slow and wasteful, espectially on slower network.

Fix this by reuse a hyper client which shares a connection pool.

I'm able to lower /r/popular loading time from 5s to 1.5s on my machine.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,15 +1,23 @@
 use cached::proc_macro::cached;
 use futures_lite::{future::Boxed, FutureExt};
-use hyper::{body, body::Buf, client, header, Body, Method, Request, Response, Uri};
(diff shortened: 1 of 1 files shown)


sigaloid (approved):
Looks great to me, thank you for your contribution!

Daniel-Valentine (approved):
Looks good. I do worry that we're becoming too reliant on mutable statics . As part of addressing #702 and #718 I'm reworking how we manage the config in memory so that it doesn't rely on a `once_cell::sync::Lazy` and so that a `crate::config::Config` object is shared between function calls; I'm thinking we should probably do something similar with our Hyper connection pool.

wsy2220:
Yeah. I don't like globals either. An intrusive refactor is too much for my first PR though.

Issue #716, opened 2023-02-11
Reduce the prominence of the instance info button


blackpill0w:
## Describe the feature you would like to be implemented
Changing the appearance of the button.

## Describe alternatives you've considered
Maybe just text saying information/about in the very bottom or maybe add it to the top bar.

mikupls:
Making the button less prominent (https://github.com/libreddit/libreddit/commit/37f71c48d1b1e4636e7a07bb6398f6829fa91de8) was a step in the right direction, and I also vote for going further. Instance information is not something a normal user cares about and it is currently very prominently displayed (especially in threads with few or no comments).

IMO the same goes for the "code" button on the top right.

I'd like to see both move to the bottom of the page (see e.g. the small links at the bottom of GitHub itself).
Or consider moving them both into the top bar (IMO the worse option). But at least use a consistent place for developer-focused information.
The libreddit version from the top left could also go there (the only reason to check it is for devs to see if you are on the latest version, right?)

If this idea is generally okay (moving dev info to the bottom), I'm happy to give it a try.

PR #728, opened 2023-02-19
Move unimportant links to footer

spikecodes:
Moves version number and code link to a minimal footer at the bottom middle of each page. This allows version number to be seen on mobile. Also reduced prominence of instance info button to address https://github.com/libreddit/libreddit/issues/716

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -32,7 +32,6 @@
 		<nav>
 			<div id="logo">
 				<a id="libreddit" href="/"><span id="lib">lib</span><span id="reddit">reddit.</span></a>
-				<span id="version">v{{ env!("CARGO_PKG_VERSION") }}</span>
 				{% block subscriptions %}{% endblock %}
 			</div>
 			{% block search %}{% endblock %}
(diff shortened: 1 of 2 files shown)


Daniel-Valentine (approved):
Looks good to me. @sigaloid, thoughts?

templates/base.html
```diff
@@ -71,10 +63,16 @@
 			{% endblock %}
 		</main>
 		{% endblock %}
+
+		<!-- FOOTER -->
 		{% block footer %}
 			<footer>
-				<div class="info-button">
-					<a href="/info" title="View instance information">&#x24D8;</a>
+				<p id="version">v{{ env!("CARGO_PKG_VERSION") }}</p>
```
Daniel-Valentine, line 70: Gave this some thought and I think we should wait on merging this until after I rework [`generate-instances-json.sh` in libreddit/libreddit-instances](https://github.com/libreddit/libreddit-instances/blob/3eb6c9124fa091f2fd5e550ed6fa89795d20b7a7/generate-instances-json.sh) to start using the instance info endpoint. If we merge this in now, [the version scraper will fail to pick up the version](https://github.com/libreddit/libreddit-instances/blob/3eb6c9124fa091f2fd5e550ed6fa89795d20b7a7/generate-instances-json.sh#L480) which will cause the script to reject the instance because the HTML tag with id `version` is now a `p` rather than a `span`.

Alternatively, we could change the script to accept either a `span` or a `p`, but really, we should probably be using that instance info endpoint to get the version, along with whatever the instance settings are (e.g. SFW-only mode).

Daniel-Valentine (changes requested):

templates/base.html
Daniel-Valentine: Or, better yet, have the script just look for `id=version` since we'll only have one HTML tag with that id.

mikupls:
Tried it today, and looks good to me.

The only minor suggestion I have is to consider making the version number link to the github release page. Could be easier to find the changelog for some users. (i.e. link to https://github.com/libreddit/libreddit/releases)

templates/base.html
spikecodes: That's a good point. Yeah, let's wait on this PR until that script is reworked.

Daniel-Valentine: @spikecodes: No need. I'll adjust the script to check for `id="version"` instead and that will take care of this problem.

Daniel-Valentine: Script has been fixed. On that basis, I think this is good to go in.

Daniel-Valentine:
@mikupls: This is a good idea, but I think we should wait to do that until [`generate-instances-json.sh` to get the version number from the `/info` endpoint, rather than scraping an element with `id="version"` which it is doing presently](https://github.com/libreddit/libreddit-instances/blob/8f4dfbe2523f117cc0390ebbbb1c0e60727032e0/generate-instances-json.sh#L480).

Daniel-Valentine (approved):

Issue #560, opened 2022-07-27
✨ Feature parity: Subscriptions saved across devices/browsers (self hosted instance only)


ozzeruk:
## How does this feature work on Reddit?

I can log into Reddit from any browser or device and it 'has remembered' which subreddits I read (all 200 odd of them).

## Describe how this could be implemented into Libreddit

Currently, Libreddit uses cookies to save which subreddits are subscribed to. I self-host Libreddit, and the same information could be in a simple text file inside the Docker container where I'm running Libreddit.

## Describe alternatives you've considered

One alternative is to subscribe to each new subreddit I discover and want to follow on each device I use. This isn't ideal and with 3 computers and a couple of mobile devices it isn't practical.

## Additional context / screenshot

I appreciate this feature doesn't make sense on a public Libreddit instance.
However, this feature makes complete sense on a privately self-hosted Libreddit instance.
I believe this feature should be very straightforward to implement as effectively we're just pointing Libreddit to a cookie text file inside the docker container rather than one held by the browser.
Without the feature, if I move to a different computer and view my private Libreddit instance, I lose everything and have to resubscribe to everything.

xrfmkrh:
you can already use this feature with environment variable.
like this `LIBREDDIT_DEFAULT_SUBSCRIPTIONS=askreddit+selfhosted`

u175:
Hi, is there an environment variable for subscribing to users?

spikecodes:
Hi @ozzeruk, as shared by @xrfmkrh and @nemesio65 [here](https://github.com/libreddit/libreddit/issues/359#issuecomment-989179035), this is possible with the `LIBREDDIT_DEFAULT_SUBSCRIPTIONS` environment variable. Only downside is you have to change the env var to change your subscriptions (the default seems to override what you subscribe to in the web interface).

@u175 Yes, a subscription to `u/urfaselol` is the same as subscribing to `r/u_urfaselol` so simply prepend `u_` to your user subscriptions in the environment variable:
```bash
LIBREDDIT_DEFAULT_SUBSCRIPTIONS=askreddit+selfhosted+u_urfaselol
```

PR #732, opened 2023-02-26
Fix default subscriptions

### LINKED ISSUE #689: LIBREDDIT_DEFAULT_SUBSCRIPTIONS setting is no longer read

sigaloid:
Cut-down #706. Fixes #689

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,8 @@ Assign a default value for each user-modifiable setting by passing environment v
 | `USE_HLS`               | `["on", "off"]`                                                                                     | `off`         |
 | `HIDE_HLS_NOTIFICATION` | `["on", "off"]`                                                                                     | `off`         |
 | `AUTOPLAY_VIDEOS`       | `["on", "off"]`                                                                                     | `off`         |
-| `HIDE_AWARDS`           | `["on", "off"]`                                                                                     | `off`
+| `HIDE_AWARDS`           | `["on", "off"]`                                                                                     | `off`         |
+| `SUBSCRIPTIONS`         | Array of subreddit names (`["sub1", "sub2"]`)                                                       | `[]`          | 
 
 You can also configure Libreddit with a configuration file. An example `libreddit.toml` can be found below:
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -147,3 +147,8 @@ fn test_alt_env_config_precedence() {
 	write("libreddit.toml", config_to_write).unwrap();
 	assert_eq!(get_setting("LIBREDDIT_DEFAULT_COMMENT_SORT"), Some("top".into()))
 }
+#[test]
+#[sealed_test(env = [("LIBREDDIT_DEFAULT_SUBSCRIPTIONS", "news+bestof")])]
+fn test_default_subscriptions() {
+	assert_eq!(get_setting("LIBREDDIT_DEFAULT_SUBSCRIPTIONS"), Some("news+bestof".into()));
+}
--- a/src/config.rs
+++ b/src/config.rs
@@ -52,6 +52,9 @@ pub struct Config {
 	#[serde(rename = "LIBREDDIT_DEFAULT_HIDE_AWARDS")]
 	pub(crate) default_hide_awards: Option<String>,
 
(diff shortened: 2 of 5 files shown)


src/config.rs
```diff
@@ -142,3 +147,8 @@ fn test_alt_env_config_precedence() {
 	write("libreddit.toml", config_to_write).unwrap();
 	assert_eq!(get_setting("LIBREDDIT_DEFAULT_COMMENT_SORT"), Some("top".into()))
 }
+#[test]
+#[sealed_test(env = [("LIBREDDIT_DEFAULT_SUBSCRIPTIONS", "news+bestof")])]
+fn test_default_subscriptions() {
+	assert_eq!(get_setting("LIBREDDIT_DEFAULT_SUBSCRIPTIONS"), Some("news+bestof".into()));
```
Daniel-Valentine, line 153: This appears to contradict the format of environment variable as given in the README, so I would have expected `["news", "bestof"]`.

src/instance_info.rs
```diff
@@ -153,10 +154,10 @@ impl InstanceInfo {
                 Deploy date: {}\n
                 Deploy timestamp: {}\n
                 Compile mode: {}\n
+				SFW only: {:?}\n
```
Daniel-Valentine, line 157: I agree we should move this around, but I think we should wait to do this in another deliverable.

src/main.rs
```diff
@@ -161,7 +161,7 @@ async fn main() {
 	let mut app = server::Server::new();
 
 	// Force evaluation of statics. In instance_info case, we need to evaluate
-	// the timestamp so deploy date is accurate - in config case, we need to
```
Daniel-Valentine, line 164: Surprised `rustfmt` didn't catch this.

Daniel-Valentine (changes requested):

src/config.rs
r7l: Excuse me for jumping in here. But i've asked to create this PR as i really enjoyed this option so far and not having this option around leaves anyone using this option kinda stuck on 0.26.0. It was the last version with it.

Having it `linux+gnome+news` was [how it worked before](https://github.com/libreddit/libreddit/issues/560). 

Not saying it wouldn't work as you suggest but it feels kinda odd when Libreddit itself already [supports this syntax](https://libreddit.spike.codes/r/linux+gnome+news) and also Reddit works this way [as you can see here](https://www.reddit.com/r/linux+gnome+news).

Daniel-Valentine: @r7l: I agree that we should reintroduce this feature, but we need to make sure that our documentation is consistent and correct.

r7l: @Daniel-Valentine I might have misunderstood your comment. Sorry. Seems like it just needs a last fix to the README file instead of a rewrite to the subscription list. Good to hear.

src/instance_info.rs
sigaloid: Ah, thanks. I should have only added the diffs that were relevant but I rather removed the diffs that looked irrelevant.

src/config.rs
sigaloid: :+1: Will fix the readme in the next commit.

README.md
```diff
@@ -1,4 +1,4 @@
-# Libreddit
+
```
Daniel-Valentine: Oops. Let me fix this.

src/config.rs
Daniel-Valentine: @r7l:

> @Daniel-Valentine I might have misunderstood your comment. Sorry. Seems like it just needs a last fix to the README file instead of a rewrite to the subscription list. Good to hear.

Not an issue. Just wanted to make sure we're all on the same page.

Daniel-Valentine:
@sigaloid: I went ahead and address the README changes. This is now good to go in.

Daniel-Valentine (approved):

PR #738, opened 2023-03-03
Display previews and inline images for reddit-hosted images.

mikupls:
This version does not change the svg behaviour in the non-ireddit cases in an attempt to reduce breakage.

This is a rebased version of the >1 year old PR #390.

The main difference is that there is no modification to the html in the non-ireddit case to prevent breaking random things.
The outcome of this PR is still a highly improved user-experience with regard to subreddits that use ireddit content, while introducing minor amount of content-reflow in these cases. All subs/pages that do not make use of ireddit content will continue to be reflow-free (no change).

The underlying problem has not changed since a year ago: ireddit content does not advertise (via the json API) its dimentions, and therefore we can only know its size only we actually load the image.

Opening this PR to hopefully revive some discussion.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -163,6 +163,9 @@ impl Media {
 			gallery = GalleryMedia::parse(&data["gallery_data"]["items"], &data["media_metadata"]);
 
 			("gallery", &data["url"], None)
+		} else if data["is_reddit_media_domain"].as_bool().unwrap_or_default() && data["domain"] == "i.redd.it" {
+			// If this post contains a reddit media (image) URL.
+			("image", &data["url"], None)
 		} else {
 			// If type can't be determined, return url
 			("link", &data["url"], None)
@@ -177,6 +180,8 @@ impl Media {
 			Self {
 				url: format_url(url_val.as_str().unwrap_or_default()),
 				alt_url,
+				// Note: in the data["is_reddit_media_domain"] path above
+				// width and height will be 0.
 				width: source["width"].as_i64().unwrap_or_default(),
 				height: source["height"].as_i64().unwrap_or_default(),
 				poster: format_url(source["url"].as_str().unwrap_or_default()),
--- a/templates/utils.html
+++ b/templates/utils.html
@@ -100,6 +100,10 @@
 	{% if post.post_type == "image" %}
 	<div class="post_media_content">
 		<a href="{{ post.media.url }}" class="post_media_image" >
+			{% if post.media.height == 0 || post.media.width == 0 %}
+			<!-- i.redd.it images speical case -->
+			<img width="100%" height="100%" loading="lazy" alt="Post image" src="{{ post.media.url }}"/>
+			{% else %}
 			<svg
 				width="{{ post.media.width }}px"
 				height="{{ post.media.height }}px"
@@ -109,6 +113,7 @@
 						<img loading="lazy" alt="Post image" src="{{ post.media.url }}"/>
 					</desc>
 			</svg>
+			{% endif %}
 		</a>
 	</div>
 	{% else if post.post_type == "video" || post.post_type == "gif" %}
@@ -216,7 +221,11 @@
 	<!-- POST MEDIA/THUMBNAIL -->
 	{% if (prefs.layout.is_empty() || prefs.layout == "card") && post.post_type == "image" %}
 	<div class="post_media_content">
-		<a href="{{ post.media.url }}" class="post_media_image {% if post.media.height / post.media.width < 2 %}short{% endif %}" >
+		<a href="{{ post.media.url }}" class="post_media_image {% if post.media.height < post.media.width*2 %}short{% endif %}" >
+			{% if post.media.height == 0 || post.media.width == 0 %}
+			<!-- i.redd.it images speical case -->
+			<img width="100%" height="100%" loading="lazy" alt="Post image" src="{{ post.media.url }}"/>
+			{% else %}
 			<svg
 				{%if post.flags.nsfw && prefs.blur_nsfw=="on" %}class="post_nsfw_blur"{% endif %}
 				width="{{ post.media.width }}px"
@@ -227,6 +236,7 @@
 						<img loading="lazy" alt="Post image" src="{{ post.media.url }}"/>
 					</desc>
 			</svg>
+			{% endif %}
 		</a>
 	</div>
 	{% else if (prefs.layout.is_empty() || prefs.layout == "card") && post.post_type == "gif" %}


Daniel-Valentine:
Thanks for raising this and for your research.

I am wondering if we can mitigate the content reflow caused by ireddit content by having Libreddit do a GET of an image once and cache the dimensions. There are downsides to this approach, namely increased memory consumption and traffic to Reddit, but that would reduce if not eliminate reflow.

sigaloid:
I think that might be a good approach. There's likely to be a larger cache with more popular instances - we don't want it to be too big. Any ideas on how to mitigate runaway memory consumption?

mikupls:
Thanks for taking a look. Daniel, can you elaborate on how this approach would work? Here is what I'm guessing:

1. browser opens `r/somecoolsub`.
2. libreddit generates the html data with lines to the proxied urls (ireddit content proxy urls).
3. html content with missing image dimensions is served to the browser.
4. browser requests the same data again via proxy urls.

Proposed patch is to add a step 2b, which makes libreddit request the image data already in the backend, leading to a new step 3 in which the image dimensions are known.

I could see that working, but:
- pages with ireddit content now likely load substantially slower, because we cannot serve html content to the user before we GET possibly a large amount of images.
- code gets more complex because 
  - we conditionally fetch some image data in the backend
  - we have to calculate image dimensions 
  - we have to cache image data somehow in the backend? (is that already implemented?) Or GET ireddit images twitch (once in the backend, once when the browser requests it)

Overall, I would personally take the rare reflow cases, but others might of course disagree :)

---

I checked again how reddit itself is doing it these days. It looks like (e.g. visiting `r/anime_irl`) reddit only ever serves preview images that have dimensions in the url. However it also requires some hash in the URL which seems to only be serves from reddit's html version (not inside the json API). I did not investigate this in detail. Maybe there is some viable alternative approach in here.

While we wait for a solution, I'll maintain my patch on https://github.com/mikupls/libreddit/tree/experimental for anyone interested.

sigaloid:
> we have to cache image data somehow in the backend? (is that already implemented?) Or GET ireddit images twice (once in the backend, once when the browser requests it)

Just the dimensional data would be cached (otherwise cache would grow rapidly for popular instances). I suppose that means we get twice. That does complicate things, I see what you mean. 

I tend to agree with the reflow argument in this case...

Daniel-Valentine:
@mikupls: I did a little bit of research into this. The hope was that we could GET a small amount of data for each image, something around 20 or so bytes, to get image headers that could provide dimension info, which we could cache temporarily (we used [cached](https://lib.rs/crates/cached) extensively to create timed caches for data we get from Reddit). Unfortunately, this doesn't appear to be possible across the board; [JFIF headers](https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format), for example, will tell you the pixel density of an image but not the dimensions thereof.

That being said, I'm not sure if we would need to do two GETs for each image. Maybe there is a way we can harness the proxy to get the dimensions from the image and then supply that to the HTTP handler for the Reddit post route somehow? I will continue to explore.

Personally though I'm not averse to reflow if it's neither common nor major. Since @spikecodes was the one to raise the issue first, I'll let him offer an opinion.

mikupls:
> Personally though I'm not averse to reflow if it's neither common nor major. Since @spikecodes was the one to raise the issue first, I'll let him offer an opinion.

I encourage everyone to just try the patch :) I've been running it for the last weeks (my fork), and found no problems so far (e.g. distorted images, like in the old PR). Granted that I used a stable internet connection (not mobile) most of the time, where reflow is a much smaller problem.
Spot-checking many popular subs makes me think that ireddit content is not that common, i.e. many subs are unaffected by this change. It seems like there is a (small?) group of subs that rely heavily on ireddit content (such as r/anime_irl) that will be impacted by this change (both positively and negatively).

The most realistic (reflow-free) alternative I can see is still somehow making use of the preview images (like reddit itself does it).

mikupls:
@oifj34f34f Thanks for showing interest in this patch. I think development in this repo is stuck/abandoned at the moment, so I don't think this will get merged anytime soon.

If you want to self-host, I maintain a version with my patches (mainly this one and some other small stuff) on https://github.com/mikupls/libreddit/tree/experimental

azizLIGHT:
> @oifj34f34f Thanks for showing interest in this patch. I think development in this repo is stuck/abandoned at the moment, so I don't think this will get merged anytime soon.
> 
> If you want to self-host, I maintain a version with my patches (mainly this one and some other small stuff) on https://github.com/mikupls/libreddit/tree/experimental

How can I use your specific fork? I have docker but I don't see a image

mikupls:
@azizLIGHT hey :wave: 

Not sure which platform you are on. On a Linux OS this should roughly work:
```
git clone https://github.com/mikupls/libreddit
cd libreddit
git checkout experimental
docker build -f Dockerfile
# wait for 2-3 minutes
# it will output some $ID
docker run -d -p 127.0.0.1:8080:8080 $ID
# go to http://127.0.0.1:8080 in your browser and it should work.
# at the bottom it should say `v0.30.1-mikupls-3`.
```

Let me know if you need more help. If you are looking for any specific patch to be added, I can also consider it.

azizLIGHT:
> Let me know if you need more help. If you are looking for any specific patch to be added, I can also consider it.

I've been using this tree happily after using your instructions. Thanks for that. I seem to recall in previous comments/PR you had one version that was even more inlined/previewed and this is the toned down version because of the resulting display issues. You had said that you didn't experience those issues that much. I'd like to try that version if it's possible?

I'm not sure how to ask you that so I replied here

skydiver1980 (approved):

oifj34f34f:
@mikupls I tried

```
git clone https://github.com/mikupls/libreddit
cd libreddit
git checkout experimental
cargo build --release
```
And then `./libreddit`, but the images still don't show up



Am I doing something wrong?

azizLIGHT:
> But the images still don't show up
> 
> 
> 
> Am I doing something wrong?

# Images in comments of a post
As far as I understand, that is a comment on a post page. I'm not sure if this version by mikupls acts on comments. You can try imagus mod ([Firefox](https://addons.mozilla.org/en-US/firefox/addon/imagus-mod/) / [Chrome](https://github.com/TheFantasticWarrior/chrome-extension-imagus)) for images in comments. Works well for subreddits whose posts are very image heavy, like /r/photoshoprequests or /r/photoshopbattles where all top-level comments are images. Helps on other sites too

image

https://github.com/libreddit/libreddit/assets/8013867/ae74623c-6566-435f-a834-0b46504035a3

# Images in the post itself
The [old PR](https://github.com/libreddit/libreddit/pull/390#issuecomment-1008437107) is talking about the post itself and is supposed to embed that.
These are the pics from the old PR.
Before (right side) and After (left side):


> when browsing:
when browsing


> when inside the post:
when inside the post

I have mikupls patch working and embedding images on the post just like the pic in the old PR's after pics:
[when browsing:](http://192.168.0.21:28080/r/linux/search?q=curious+history+experience&restrict_sr=on):
image

[when inside the post:](http://192.168.0.21:28080/r/linux/comments/rzphey/im_curious_about_the_history_of_in_box_linux_os/)
image

oifj34f34f:
@azizLIGHT Thanks for the clarification. I mistakenly thought it was a PR to fix the comment images.. xD

sigaloid:
Hi @mikupls, I'm working on a fork right now, I merged this into it. Just to confirm, as is, this will result in cache growing more with popularity as it will cache images, right?

PR #740, opened 2023-03-05
show the count of 'more replies'.

### LINKED ISSUE #364: Show the quantity of 'more replies'
## How does this feature work on Reddit?

At the bottom of a top level comment, reddit shows a link to see more replies, while detailing the number of replies:
* `38 more replies` on new reddit
* `load more comments (45 replies)` on old reddit
* `load more comments (45 replies)` on i.reddit.com / reddit compact
* `load more comments (45)` on teddit.net
* `39 more replies` on third party android client Boost

Clicking the link loads a new page with some more comments. Indeed, the link to see more replies exists in nested comments at any level and the behavior is the same there as well. The number of replies helps the user determine if a click is warranted or not and whether to incur a page load or not, especially for those with limited resources: patience, processing power, limited data, etc.

## Describe the implementation into Libreddit

Currently, Libreddit shows a link `→ More replies` without any numerical indications.

Instead, Libreddit could add the number of replies to the link:  `→ 38 more replies`

Old reddit, new reddit, compact reddit, as well as teddit.net and an android client all use this behavior

## Additional context

Libreddit:
 Libreddit

Reddit redesign:
Reddit redesign

Old Reddit:
Old Reddit

Compact/i Reddit:
Compact Reddit

Teddit.net:
teddit.net

mikupls:
Fixes #364. Only took 2 years :sweat_smile: 

I only realized how useful this feature is after implementing it and using it for a bit. Thanks for creating the request.

---

Example:

pic

### DIFF
--- a/src/post.rs
+++ b/src/post.rs
@@ -107,6 +107,13 @@ fn parse_comments(json: &serde_json::Value, post_link: &str, post_author: &str, 
 
 			let score = data["score"].as_i64().unwrap_or(0);
 
+			// The JSON API only provides comments up to some threshold.
+			// Further comments have to be loaded by subsequent requests.
+			// The "kind" value will be "more" and the "count"
+			// shows how many more (sub-)comments exist in the respective nesting level.
+			// Note that in certain (seemingly random) cases, the count is simply wrong.
+			let more_count = data["count"].as_i64().unwrap_or(0);
+
 			// If this comment contains replies, handle those too
 			let replies: Vec<Comment> = if data["replies"].is_object() {
 				parse_comments(&data["replies"], post_link, post_author, highlighted_comment, filters, req)
@@ -177,6 +184,7 @@ fn parse_comments(json: &serde_json::Value, post_link: &str, post_author: &str, 
 				awards,
 				collapsed,
 				is_filtered,
+				more_count,
 				prefs: Preferences::new(req),
 			}
 		})
--- a/src/post.rs
+++ b/src/post.rs
@@ -112,7 +112,7 @@ fn parse_comments(json: &serde_json::Value, post_link: &str, post_author: &str, 
 			// The "kind" value will be "more" and the "count"
 			// shows how many more (sub-)comments exist in the respective nesting level.
 			// Note that in certain (seemingly random) cases, the count is simply wrong.
-			let more_count = data["count"].as_i64().unwrap_or(0);
+			let more_count = data["count"].as_i64().unwrap_or_default();
 
 			// If this comment contains replies, handle those too
 			let replies: Vec<Comment> = if data["replies"].is_object() {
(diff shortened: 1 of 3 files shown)


mikupls:
Of course I find the first weird case shortly after creating the PR :)

pic2

Not sure what that is about yet (or if it should block this PR).

src/post.rs
```diff
@@ -107,6 +107,12 @@ fn parse_comments(json: &serde_json::Value, post_link: &str, post_author: &str,
 
 			let score = data["score"].as_i64().unwrap_or(0);
 
+			// The API only provides comments up to some threshold.
+			// Further comments have to be loaded by subsequent requests.
+			// The "kind" value will be "more" and the "count"
+			// shows how many more comments exist in the current nesting level.
+			let more_count = data["count"].as_i64().unwrap_or_default().to_string();
```
sigaloid, line 114: Do we need this to be a string value? I'm pretty sure we can format as an i64 directly into askama. Not that it's a huge or expensive allocation, but if the added layer of misdirection is unnecessary, it's always a good idea to reduce it!

sigaloid:
Yep, that edge case does exist. Reddit will flat out lie to you about the replies - happens to the site, JSON api, even third party apps. 

Is it possible that we can cover this case by checking the length of the children array?

Thanks for the PR btw!

sigaloid (approved):
Besides this, LGTM!

src/post.rs
mikupls: Adjusted it, thanks.

mikupls:
> Yep, that edge case does exist. Reddit will flat out lie to you about the replies - happens to the site, JSON api, even third party apps.
> 
> Is it possible that we can cover this case by checking the length of the children array?
> 
> Thanks for the PR btw!

Thanks for the quick review!

From what I can tell, the JSON API response does not contain the children, just a "more" count. So "verifying" the count would actually involve the backend doing another request and checking against that response. Which would make everything slower.

I propose merging this as-is for now, and as we find more edge-cases (possibly reported by more users using this feature), we can see if some optimizations are possible.

In general I have a feeling that deleted comments play a role in reddit's "miscalculations". But hard to know for sure.

src/post.rs
```diff
@@ -107,6 +107,13 @@ fn parse_comments(json: &serde_json::Value, post_link: &str, post_author: &str,
 
 			let score = data["score"].as_i64().unwrap_or(0);
 
+			// The JSON API only provides comments up to some threshold.
+			// Further comments have to be loaded by subsequent requests.
+			// The "kind" value will be "more" and the "count"
+			// shows how many more (sub-)comments exist in the respective nesting level.
+			// Note that in certain (seemingly random) cases, the count is simply wrong.
+			let more_count = data["count"].as_i64().unwrap_or(0);
```
sigaloid, line 115: I think this will work:

```suggestion
			let more_count = data["count"].as_i64().unwrap_or_default();
```

PR #741, opened 2023-03-05
fix build error on windows

o69mar:
i expected libreddit to not work on windows, but when i installed it using cargo i got a build error, i fixed it and recompiled libreddit, ran it and it works just fine.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,13 @@
 use std::{
-	os::unix::process::ExitStatusExt,
-	os::windows::process::ExitStatusExt,
 	process::{Command, ExitStatus, Output},
 };
+
+#[cfg(not(target_os = "windows"))]
+use std::os::unix::process::ExitStatusExt;
+
+#[cfg(target_os = "windows")]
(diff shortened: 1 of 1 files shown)


sigaloid (commented):
This causes an error - can't import `os::windows` on linux. Perhaps something like this:

```rust
use std::{
	process::{Command, ExitStatus, Output},
};

#[cfg(not(target_os = "windows"))]
use std::os::unix::process::ExitStatusExt;

#[cfg(target_os = "windows")]
use std::os::windows::process::ExitStatusExt;
```

o69mar:
> This causes an error - can't import `os::windows` on linux. Perhaps something like this:
> 
> ```rust
> use std::{
> 	process::{Command, ExitStatus, Output},
> };
> 
> #[cfg(not(target_os = "windows"))]
> use std::os::unix::process::ExitStatusExt;
> 
> #[cfg(target_os = "windows")]
> use std::os::windows::process::ExitStatusExt;
> ```
> 
> Otherwise, this looks great, thank you for the first contribution smile crab!

fixed it, yw

PR #745, opened 2023-03-08
Store subreddit subscriptions and filter in lower-case

### LINKED ISSUE #730: 🐛 Bug Report: Cannot Unsubscribe from banned subreddits
## Describe the bug

Attempting to remove a subreddit that is banned from my subscriptions list as it is redundant and not needed anymore. Met with a screen that says 'Banned" and subreddit stays on my subscription lists.

## Steps to reproduce the bug

1. Go to settings page.
2. Click on unsubscribe from banned subreddit.
3. met with "Banned" screen and subreddit not removed from lists.

## What's the expected behavior?
 
Subreddit should remove from subscription lists.

staticssleever668:
This allows us to prevent an API call in subscriptions_filters(). This in turn fixes unsubscribing from banned subreddits (fixes #730).
As a side effect, in the settings page subreddit names are also inlower-case, instead of being in their correct casing.


src/subreddit.rs
```diff
@@ -210,37 +210,8 @@ pub async fn subscriptions_filters(req: Request<Body>) -> Result<Response<Body>,
 	let mut sub_list = preferences.subscriptions;
 	let mut filters = preferences.filters;
 
-	// Retrieve list of posts for these subreddits to extract display names
-	let posts = json(format!("/r/{}/hot.json?raw_json=1", sub), true).await?;
```
staticssleever668, line 214: Maybe it would be a better idea to handle the error here instead of propagating it, but still, this allows to save an API call. What do you think?

src/server.rs
```diff
@@ -379,7 +379,7 @@ fn determine_compressor(accept_encoding: String) -> Option<CompressionType> {
 
 	// This loop reads the requested compressors and keeps track of whichever
 	// one has the highest priority per our heuristic.
-	for val in accept_encoding.to_string().split(',') {
+	for val in accept_encoding.split(',') {
```
Daniel-Valentine, line 382: I'm not opposed to this, but in the future, let's raise small fixes like these in a separate deliverable.

src/utils.rs
```diff
@@ -530,21 +532,21 @@ impl Preferences {
...
-			use_hls: setting(&req, "use_hls"),
-			hide_hls_notification: setting(&req, "hide_hls_notification"),
-			autoplay_videos: setting(&req, "autoplay_videos"),
-			disable_visit_reddit_confirmation: setting(&req, "disable_visit_reddit_confirmation"),
-			comment_sort: setting(&req, "comment_sort"),
-			post_sort: setting(&req, "post_sort"),
-			subscriptions: setting(&req, "subscriptions").split('+').map(String::from).filter(|s| !s.is_empty()).collect(),
-			filters: setting(&req, "filters").split('+').map(String::from).filter(|s| !s.is_empty()).collect(),
-			hide_awards: setting(&req, "hide_awards"),
+			theme: setting(req, "theme"),
```
Daniel-Valentine, line 535: Vis-a-vis these `&req` to `req` changes, see my comment on src/server.rs:382.

src/subreddit.rs
Daniel-Valentine: I'll leave it up to @spikecodes and/or @sigaloid. In my opinion, though, we should strive for a consistent user experience across bugfixes whenever possible, so I would vote for handling the error and perhaps annotating the banned subreddit with something like a ` (banned)` prefix, e.g. `nonewnormal (banned)`.

Daniel-Valentine (changes requested):

src/server.rs
staticssleever668: You are right, this one is outside of the scope of this PR, not sure why I staged that.

src/utils.rs
staticssleever668: TBH I think it does make a lot of sense to have this change together with the fix as the fix directly modifies the code in question.

staticssleever668:
Not that I look at this again, I do agree this is a sloppy fix, I'll try to remake this.

PR #748, opened 2023-03-10
Update README.md

xatier:
* Remove duplicated config

Was accidentally introduced in  https://github.com/libreddit/libreddit/commit/412ce8f1f3ab7ef91f8a40182aee2a039bd7783b


xatier:
@sigaloid, please kindly take a look. :pray:

xatier:
I actually wonder if we should keep the same order as in  https://github.com/libreddit/libreddit/blob/master/src/config.rs#L22?

I.e., swap POST_SORT and COMMENT_SORT, and putting HIDE_AWARDS before SUBSCRIPTIONS, how do you think?

Daniel-Valentine:
I don't have strong opinions about the order in the table, so I say we accept this as-is. I apologize for messing up the table in the merge commit.

xatier:
no worries :D 
Thanks for merging this!

xatier:
@Daniel-Valentine created another PR for the rearrange. Thanks! https://github.com/libreddit/libreddit/pull/749

Issue #641, opened 2022-11-18
🐛 Bug Report: aarch64 Docker builds have been failing for 6 months


tazz4843:
Latest output logs: https://github.com/libreddit/libreddit/actions/runs/3466848096/jobs/5791127262.

Exit code 137 makes me think something was attacked by SIGKILL (137-128 = 9: SIGKILL), which is very likely the OOM reaper. Possible fix would be decreasing the concurrency for these builds (`-j 1` perhaps?). Not 100% sure as the logs are next to useless, so other input here is a good idea.

Daniel-Valentine:
@spikecodes: I know you were looking into this a while back. Any update on this front?

Daniel-Valentine:
(temporarily reopening; will close if/when fix is confirmed)

Daniel-Valentine:
Per [this discussion](https://users.rust-lang.org/t/cargo-uses-too-much-memory-being-run-in-qemu/76531) on the Rust language forum, it appears Cargo's git library is consuming too many resources, specifically RAM, for Linux's liking. The solution is to tell cargo to use git the binary rather than a library.

[A build is underway](https://github.com/libreddit/libreddit/actions/runs/3527659829/jobs/5916958140). Hopefully, this goes through. If not, we may need to instruct cargo to execute only one job at a time. Unfortunately, this means the ARM Docker build will take even more time than it does now, which, given the VMs limited resources, is already a lot (the build started 30 minutes ago as of the time I wrote this comment).

I suspect this is a manifestation of rust-lang/cargo#10583. The config fix I mentioned above is also offered as a suggestion in [one of the comments for that issue](https://github.com/rust-lang/cargo/issues/10583#issuecomment-1129997984).

Daniel-Valentine:
The build passed. I'm going to consider this fixed, but I'll keep my eye out to make sure we don't see this again. If we do, we may have to impose a cargo job limit of 1 as I said earlier.

PR #751, opened 2023-03-13
optimize arm dockerfile

master-hax:
cache the rust dependencies in their own docker layer, so they will be reused for subsequent builds as long as the `Cargo.toml` or `Cargo.lock` files are not updated.

### DIFF
--- a/Dockerfile.arm
+++ b/Dockerfile.arm
@@ -7,6 +7,10 @@ RUN apk add --no-cache g++ git
 
 WORKDIR /usr/src/libreddit
 
+# cache dependencies in their own layer
+COPY Cargo.lock Cargo.toml .
+RUN mkdir src && echo "fn main() {}" > src/main.rs && cargo install --config net.git-fetch-with-cli=true --path . && rm -rf ./src
+
 COPY . .
 
 # net.git-fetch-with-cli is specified in order to prevent a potential OOM kill


sigaloid:
Rather than using the git CLI, perhaps we could switch to using the HTTP lighter index? It's on nightly currently but can be enabled with a flag/env var. This might further optimize the compile time.

master-hax:
Hey @sigaloid thanks for taking a look at my PR.

What is "HTTP lighter index"? Is it the same thing as the Cargo [sparse-registry](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#sparse-registry)?

sigaloid:
Yep, that's exactly what I meant. Is that something you'd want to add? I just figured it was related to optimization and the index, and whatnot. No worries if not!

master-hax:
i tried using `FROM rustlang/rust:nightly-alpine AS builder` as well as `ENV CARGO_UNSTABLE_SPARSE_REGISTRY=true` in the Dockerfile as described [here](https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html) but did not notice any difference in speed.

not sure if it worked? fwiw i am building on a raspberry pi 4b 8gb so everything takes a really long time to compile. https://github.com/libreddit/libreddit/issues/641 seems to have more context on why we decided to use the git CLI

master-hax:
i think we can move forward without the sparse registry support

sigaloid:
Apologies for the delay. I'm going to approve this, thank you for the contribution!

PR #753, opened 2023-03-14
add link to reddit status page.

mikupls:
Reddit is down. Good opportunity to make the error page slightly more nice to use.

Preview:
pic123

### DIFF
--- a/templates/error.html
+++ b/templates/error.html
@@ -4,6 +4,8 @@
 {% block content %}
 	<div id="error">
 		<h1>{{ msg }}</h1>
+		<h3><a href="https://www.redditstatus.com/">Reddit Status</a></h3>
+		<br>
 		<h3>Head back <a href="/">home</a>?</h3>
 	</div>
 {% endblock %}


mikupls:
I realize the error page is a bit overloaded in use (i.e. there are many paths that lead to the error page). Many of these paths point to some problem with reddit though, and it didn't feel worth to refactor everything for a small link that is likely useful in most cases.

sigaloid (approved):
Looks great to me, thanks for the PR!

templates/error.html
```diff
@@ -4,6 +4,8 @@
 {% block content %}
 	<div id="error">
 		<h1>{{ msg }}</h1>
+		<h3><a href="https://www.redditstatus.com/">Reddit Status</a></h3>
+		<br>
```
Daniel-Valentine, line 8:
```suggestion
		<br />
```

Daniel-Valentine (changes requested):

skydiver1980 (approved):

PR #760, opened 2023-03-17
harden docker-compose.yml

kuanhulio:
`user: nobody`: the least privileged account.
`read_only: true`: this container doesn't write anything to the filesystem, this removes a vector.
`security_opt`: disallows the container to grab more privileges.
`cap_drop`: this container doesn't need any capabilities, drop them.
`networks`: put `libreddit` into its own network so it cannot see other containers by default.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,18 @@ services:
     container_name: "libreddit"
     ports:
       - 8080:8080
+    user: nobody
+    read_only: true
+    security_opt:
+      - no-new-privileges: true
+    cap_drop:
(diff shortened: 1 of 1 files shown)


Daniel-Valentine:
Looks good to me. @sigaloid, @spikecodes: Thoughts?

kuanhulio:
Doing a cursory glance, I also see #441, I'm not sure whether or not that works, but the `seccomp` is also a good idea, haven't personally tested his configuration however.

Daniel-Valentine (approved):

PR #761, opened 2023-03-17
switch to a distroless image; update docker-compose to reflect it.

kuanhulio:


kuanhulio:
Sorry, it just completely blanked on my mind that distroless containers exist. Should've been included in #760, that's my bad.

sigaloid:
Hi, I like this change. Do you have any specific numbers on the size/time saved on the image?

kuanhulio:
Building from the main branch: `14.77 MB`
Building from this PR: `10.17 MB`

Not an insane difference as the main branch is already really optimized, but it's an improvement nonetheless.

Buildtimes are the same, however. I don't know enough Rust to optimize the buildtime anyway without some level of jank-ness.

sigaloid:
Alright, great, this change is good with me. @Daniel-Valentine @spikecodes thoughts?

kuanhulio:
In theory, we could also do a `scratch` build.

Building the `scratch` part results in a `7.71 MB` build.

PR #769, opened 2023-03-22
Allow bypassing nsfw gate for posts

### LINKED ISSUE #758: 💡 Feature request:  Disabling the NSFW warning without enabling the Show NSFW filter.
After the update, which added a warning when viewing pages tagged NSFW, such as bookmarked pages or user profiles, the user would have to enable the Show NSFW posts in the settings, save the settings and reload the page then disable it again, which to me is rather cumbersome to just view a page and I often forget to disable it.

## Describe the feature you would like to be implemented
Disabling the NSFW warning without enabling the Show NSFW filter.

I decided to use older instances of libreddit, but I'm afraid that using older instances isn't sustainable, since those could be updated sooner or later. I also like the hide awards option of the recent version, since it makes things less distracting, so I decided to make this suggestion.

Both of these instances have their Show NSFW posts option disabled, but with the older one, I can still see my bookmarked page and go to my Home without seeing other NSFW posts.
libreddit v25
libreddit v30

gmnsii:
Closes #758
On instances that are not sfw-only, the nsfw gate for posts can now be bypassed without changing preferences.

### DIFF
--- a/templates/nsfwlanding.html
+++ b/templates/nsfwlanding.html
@@ -19,10 +19,12 @@
         {% if crate::utils::sfw_only() %}
         This instance of Libreddit is SFW-only.</p>
         {% else %}
-        Enable "Show NSFW posts" in <a href="/settings">settings</a> to view this {% if res_type == crate::utils::ResourceType::Subreddit %}subreddit{% else if res_type == crate::utils::ResourceType::User %}user's posts or comments{% else if res_type == crate::utils::ResourceType::Post %}post{% endif %}.
+        Enable "Show NSFW posts" in <a href="/settings">settings</a> to view this {% if res_type == crate::utils::ResourceType::Subreddit %}subreddit{% else if res_type == crate::utils::ResourceType::User %}user's posts or comments{% else if res_type == crate::utils::ResourceType::Post %}post{% endif %}. <br>
+        {% if res_type == crate::utils::ResourceType::Post %} You can also bypass this gate by clicking on this <a href="{{url}}&bypass_nsfw_landing">link</a>.{% endif %}
         {% endif %}
     </p>
 </div>
 {% endblock %}
 {% block footer %}
 {% endblock %}
+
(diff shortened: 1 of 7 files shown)


templates/nsfwlanding.html
```diff
@@ -19,10 +19,12 @@ <h1>
         {% if crate::utils::sfw_only() %}
         This instance of Libreddit is SFW-only.</p>
         {% else %}
-        Enable "Show NSFW posts" in <a href="/settings">settings</a> to view this {% if res_type == crate::utils::ResourceType::Subreddit %}subreddit{% else if res_type == crate::utils::ResourceType::User %}user's posts or comments{% else if res_type == crate::utils::ResourceType::Post %}post{% endif %}.
+        Enable "Show NSFW posts" in <a href="/settings">settings</a> to view this {% if res_type == crate::utils::ResourceType::Subreddit %}subreddit{% else if res_type == crate::utils::ResourceType::User %}user's posts or comments{% else if res_type == crate::utils::ResourceType::Post %}post{% endif %}. <br>
+        {% if res_type == crate::utils::ResourceType::Post %} You can also bypass this gate by clicking on this <a href="{{url}}&bypass_nsfw_landing">link</a>.{% endif %}
```
sigaloid, line 23: I would mention that it's only temporarily bypassing this gate.

sigaloid (changes requested):
Thanks for the contribution! The code looks good for the most part, I'll wait on some other opinions on whether or not this should go in. Also please run cargo fmt on this code! Thanks :)

templates/nsfwlanding.html
gmnsii: Thanks for the heads-up, I changed it.

gmnsii:
I changed the bypass message to be clearer and formatted the rust code.

sigaloid:
I'm all good with this change, if @Daniel-Valentine or @spikecodes have any objections let me know, otherwise I'll merge this soon. Thank you for the contribution!

sigaloid:
Fixed merge conflicts. Merging, thanks again for the contribution!

PR #770, opened 2023-03-23
fix(style): fit footer width to body size

### LINKED ISSUE #750: Footer creates horizonal scroll bar
## Describe the bug
Footer creates horizonal scroll bar

The footer seems a little too wide. The scroll bar is gone if I remove the footer.

I'm on linux chrome 111.0.5563.64 

2023-03-13-004028_2342x2007_scrot

invakid404:
Closes #750.

This patch makes the footer's width 100% of the body's width to ensure that it doesn't stick out. This gets rid of the horizontal scroll bar.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -293,7 +293,7 @@ body > footer {
 	display: flex;
 	justify-content: center;
 	align-items: center;
-	width: 100vw;
+	width: 100%;
 	background: var(--post);
 	position: absolute;
 	bottom: 0;


invakid404:
BTW forgot to mention that Firefox works fine before this patch, the issue appears to be a Chromium-only thing. This patch only makes a difference on Chromium-based browsers, there's no difference on Firefox.

miclgael (approved):
LGTM - Will fix https://github.com/libreddit/libreddit/issues/750

sigaloid:
Thanks for the fix!

PR #778, opened 2023-04-06
added comments to libreddit.service and .conf

Akanksh12:
Added comments to contrib/libreddit.conf so it is easier and more convenient to use.


sigaloid:
Hi, just curious why you commented the default IP and port on contrib/libreddit.service? I think it would be good to leave it as the correct default port.

Akanksh12:
I commented out the default IP and port in libreddit.service as they're already there in contrib/libreddit.conf. I agree that leaving it on the default port would be good, can you tell me how to fix that? I'm new to contributing. Should I make another PR?

Akanksh12:
Hi, I've changed the default port to 12345

sigaloid:
Thanks for the contribution! Merging.

PR #802, opened 2023-05-11
Clarify JavaScript usage

oifj34f34f:
Libreddit actually uses a bit of JavaScript


miclgael (approved):
LGTM, good pick up imo

PR #811, opened 2023-06-03
Add option for Pushshift API site

Tokarak:
^_^
Revival of #706. Fixes #682.

PR #812, opened 2023-06-03
Rebase #811

Tokarak:


sigaloid:
Thanks for the fix, merging!

Fixes #682

PR #814, opened 2023-06-03
Bump dependencies + cargo update

Tokarak:
Why?

During recent debugging, I had to use `toml::Table` in toml 0.7.4. But toml 0.5.10 had Table in `toml::value::Table`. I figured since toml is two major versions behind, it's time to update. Fortunately I don't think there are any breaking changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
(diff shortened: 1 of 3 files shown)


Tokarak:
My IDE was complaining about the markdown tables in README.md. It's probably not good to squash these commits together.

sigaloid:
Thanks for the contribution!

Issue #818, opened 2023-06-04
Investigate alternate private API's that new Reddit uses


sigaloid:
Related to #785 

We need to be prepared in a month for the Reddit API changes. We can begin by looking into private API's that the web app uses. (We don't need to resort to webscraping yet)

hogseedy:
Main Reddit GQL schema as used by the Android app `(id, name, query)`: https://gist.github.com/hogseedy/a83a632aedbc813c4dbc0353422efd5f

This dump doesn't have the object descriptions, because they're scattered all around the apk, but it's possible to extract all of them algorithmically. I'll work on it.

Example code using reddit app anonymous mode/native authentication to query GQL (acquired tokens also work on public apis, native app uses them a lot):
https://gist.github.com/hogseedy/b149c5f1ad1b628ba00556c7d4a898f8


Hope it helps!

sigaloid:
Absolutely amazing. Thank you so much! This will help **enormously.** I'm going to see what I can do with this.

luphoria:
> Main Reddit GQL schema as used by the Android app `(id, name, query)`: https://gist.github.com/hogseedy/a83a632aedbc813c4dbc0353422efd5f

Note we can also brute force a (partial) schema using something like https://github.com/nikitastupin/clairvoyance. I've used it before. But for some reason it won't work with gql.reddit.com?. `AttributeError: 'NoneType' object has no attribute 'get'` Perhaps I need to pass some headers.

You can see there used to be graphiql cached on gql.reddit.com but now it's a 403.
![image](https://github.com/libreddit/libreddit/assets/60309933/f6f922a7-b2b7-46f2-827d-9f78518d5ef9)

sigaloid:
Yeah, I got the same error when trying to use clairvoyance. I don't know if it would work anyhow, as all of the query ID's seem randomly generated and a dictionary attack won't really work. (I'm not wholly familiar with GraphQL, apologies if that's wrong)

There's also `https://graphql.kubernetes.ue1.snooguts.net` which popped up a few times in my research. Looks like another GQL endpoint of sorts. Might be of interest.

luphoria:
> I don't know if it would work anyhow, as all of the query ID's seem randomly generated and a dictionary attack won't really work. (I'm not wholly familiar with GraphQL, apologies if that's wrong)

[...]

Issue #785, opened 2023-04-19
Reddit's API changes


rubenelshof:
Yesterday Reddit announced changes to it's API.
It includes a paid option for third parties who require additional capabilities, higher usage limits, and broader usage rights.
Further it limits access to mature content & rate limiting will be enforced.
These changes will be effective on June 19, 2023.
See the post below
https://www.reddit.com/r/reddit/comments/12qwagm/an_update_regarding_reddits_api/

I would like to know how these changes might affect Libreddit.

sigaloid:
An important question we need to consider is what we do if they lock down unauthenticated requests. Is webpage scraping something that we should implement? I think it's really important to have something like Libreddit but it's a different ballgame if it's a cat and mouse game rather than just piggy-backing on the API routes which is (assumed to be) allowed to be used. 

[Web scraping _is_ legal, technically, in the USA.](https://techcrunch.com/2022/04/18/web-scraping-legal-court/) But it might not be good to stoop into the grey area especially in other countries. 

We'll have to discuss this if it happens, but it's good to ask early so we can think about it.

r7l:
Without any 3rd party apps, it will be a mess using Reddit. The site is clunky already and will only become worse in future seeing the road they are on already. Not sure if it's actually such a big deal to go the [Nitter](https://github.com/zedeus/nitter) way. Same as Libreddit they don't have login / account support and aren't using Twitter API.

sigaloid:
It might not be terrible as an alternative, but it would likely require abandoning much of the codebase and probably starting from scratch, besides the front end templates. We could also run into rate limits, anti-DDoS, bot detection, etc. Just something to consider. I'd be willing to help with that transition if it becomes necessary.

nickelswitte:
Hello,
I want to express my appreciation for this project. I am using reddit on a regular basis to find information or answers to many of my problems and really dislike using their website. LibReddit offered me a great, simple and safe way to cut away all that trash and just the see content. I hope this project will survive this API policy change and very interested in keep using it. Thank you for the work 🚀

sphars:
The developer of the well-known iOS app, Apollo, had a call with reddit themselves to discuss this. Might have some good info here with regard to API access come the change: https://www.reddit.com/r/apolloapp/comments/12ram0f/had_a_few_calls_with_reddit_today_about_the/

sigaloid:
[...]

PR #819, opened 2023-06-06
Initial PoC of spoofing Android OAuth

sigaloid:
This is a proof of concept of using the Android OAuth endpoints to create an anonymous "android user" that then authorizes all requests. Couple of caveates:
* Does not use GraphQL endpoints at all.
  * GQL is faster and less rate-limited
  * GQL is likely going to be more stable or at least harder to block - android client doesn't necessarily use the OAuth endpoints I've used here so it's theoretically possible for them to block
  * GQL would likely require at least a whole new library dedicated to GraphQL API, whereas this was as easy as auth setup + replacing www.reddit.com with oauth.reddit.com
* DOES identify itself as an Android device, in headers. 
	* DOES NOT need the HMAC secret key that was reverse engineered out of the APK - no user accounts are logging in
* DOES NOT handle re-authenticating. Looks like the token expires after 24 hours. I'd like to figure out the proper way to refresh a token so the behavior more closely matches. Not sure exactly how to do that right now. Technically, since I'm just using OAuth endpoints, I could refresh them the normal way, but perhaps there's something special the real client does? (I worry that I must match the behavior exactly or else this will just be a cat and mouse game)
* Technically, does not actually pretend to be an android client to the letter - it just asks to log in as one then authenticates into the regular OAuth endpoints. This is definitely not standard behavior - for that I'd need to implement GQL. I doubt they're doing intense checking of this though.

Related to #785 #818.

HUGE thanks to @hogseedy and [@/costco on Hacker News](https://news.ycombinator.com/item?id=36086240) for the sample code. I'm just dipping my toes in with the OAuth endpoints, haven't yet touched the GQL ones.

TODO:

[ ] Token refreshing

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,8 @@ toml = "0.7.4"
 once_cell = "1.17.0"
 serde_yaml = "0.9.16"
 build_html = "2.2.0"
+uuid = { version = "1.3.3", features = ["v4"] }
+base64 = "0.21.2"
 
 [dev-dependencies]
 lipsum = "0.9.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -483,6 +483,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "getrandom"
+version = "0.2.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c85e1d9ab2eadba7e5040d4e09cbd6d072b76a557ad64e797c2cb9d4da21d7e4"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "wasi",
+]
+
+[[package]]
 name = "globset"
 version = "0.4.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -712,6 +723,7 @@ name = "libreddit"
 version = "0.30.1"
 dependencies = [
  "askama",
+ "base64",
  "brotli",
  "build_html",
  "cached",
@@ -735,6 +747,7 @@ dependencies = [
  "tokio",
  "toml",
  "url",
+ "uuid",
 ]
 
 [[package]]
@@ -1588,6 +1601,15 @@ dependencies = [
  "form_urlencoded",
  "idna",
  "percent-encoding",
+]
+
+[[package]]
+name = "uuid"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "345444e32442451b267fc254ae85a209c64be56d2890e601a0c37ff0c3c5ecd2"
+dependencies = [
+ "getrandom",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


sigaloid:
I've added automatic token refreshing, it should now handle retrieving a new token when the expiry time elapses.

Cargo.toml
```diff
@@ -30,6 +30,12 @@ toml = "0.7.4"
 once_cell = "1.17.0"
 serde_yaml = "0.9.16"
 build_html = "2.2.0"
+uuid = { version = "1.3.3", features = ["v4"] }
+base64 = "0.21.2"
+fastrand = "1.9.0"
+log = "0.4.18"
+pretty_env_logger = "0.5.0"
+dotenvy = "0.15.7"
```
sigaloid, line 38: `uuid`: Used for spoofing the device clients
`base64`: Used for generating the HTTP Basic Auth header
`fastrand`: Used to generate random devices to spoof
`log`, `pretty_env_logger`, `dotenvy`: Used to improve logging

luphoria:
Is it planned/considered to wrap the oauth spoof around a proxy which functions like original reddit API? I think a setup like that would be better longterm to avoid hardcoding lots of endpoints, and then those endpoints changing

sigaloid:
I'm not sure what you mean by wrapping the oauth spoof around a proxy. It's currently just using the same endpoints that we used before, except with `oauth.reddit.com` as the root domain instead of `www.reddit.com`. I don't expect the OAuth API endpoints to change, as this is what real paying API consumers will use. The endpoints I use aren't really "private", just the method used to generate the tokens that makes it _appear_ it's from the official client. (That's the only really "hacky" part of this PR - spoofing of the device)

If we are to use the GraphQL queries in the future, that would definitely be an issue - I wrote a simple library that will allow the map of action titles to action ID's to be updated from the GitHub, without having to update the whole program. But I haven't started the work on that.

Tokarak:
BTW is beta testing required before the June 30th API changes?

sigaloid:
It works currently, but I've no clue whether it will work after the changes. We're basically stalled until then.

More importantly this code will unfortunately likely never be merged due to legal issues (#785). We *will* add custom API keys (it'll likely be necessary), but we cannot distribute code that automatically spoofs an official client and generates an API key, due to a myriad of trademark and copyright issues. 

The beauty of open source means we can fork whenever we want. :fork_and_knife:

skydiver1980 (approved):

maxexcloo:
Works well! Wonder if it would be worth to merge this and some other PRs (such as https://github.com/libreddit/libreddit/pull/738) into a fork (as it seems like the main repo is dead?)

sigaloid:
I'll be creating a fork like that soon. The legality is on the same level as youtube-dl as I understand it so my previous comment stands corrected.

I'll put a date on it so I force myself to. By the end of this year I'll have created a fork with this and all of the relevant backlogged PRs. I seem to be the only active maintainer so i'll keep that in mind w.r.t. maintenance.

PR #821, opened 2023-06-07
Add ability to hide subreddit panel (closes #801)

### LINKED ISSUE #801: 💡 Feature request: Add ability to hide subreddit panel

On mobile, the subreddit panel takes up nearly half the screen space: 

image

## Describe the feature you would like to be implemented

It would be great if the panel was collapsable, or if there were a toggle to disable it entirely.

fawni:
wraps the subreddit panel in a `<details>` to be able to hide it like the sidebar panel.

setting the min-width to `350px` is not ideal and does not look very good on screens with a width less than `370px`, but it is the only solution i could come up with to stop the width from changing when opening/closing the panels. if you have a better solution please suggest it.

closes #801.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -380,13 +380,17 @@ aside {
 	border-radius: 5px;
 	overflow: hidden;
 }
+#subreddit, #sidebar { min-width: 350px; }
 
 #user *, #subreddit * { text-align: center; }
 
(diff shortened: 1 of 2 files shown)


sigaloid:
Thanks for the contribution!

Issue #859, opened 2023-09-27
✨ Feature parity: follow new share links


kylrth:
## How does this feature work on Reddit?

I've started getting these new share links that point to a specific post, which Reddit probably uses to track shares. Once I open the link on Reddit, it redirects to the canonical URL for the post. Here's an example link:

```
https://reddit.com/r/Unexpected/s/hk6r0GRYOM
```

## Describe how this could be implemented into Libreddit

Right now, Libreddit shows `Error: Nothing here`. The proxy could instead follow the redirect for you and result in the final canonical URL.

## Describe alternatives you've considered

Maybe it would be smart to hide this behind an "Are you sure?" prompt, so users understand they're providing a bit of a signal to Reddit (albeit through the proxy).

PR #861, opened 2023-10-04
Handle obfuscated share links

pts314:
Adds /r/:sub/s/:id pattern in main.rs to follow obfuscated share links and strips share info. As noted in the attached issue, it could be better by adding a warning, but this at least will continue to allow things to function in the mean time

Resolves: #859

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,6 +18,7 @@ use clap::{Arg, ArgAction, Command};
 
 use futures_lite::FutureExt;
 use hyper::{header::HeaderValue, Body, Request, Response};
+use url::Url;
 
 mod client;
 use client::{canonical_path, proxy};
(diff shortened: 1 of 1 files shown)


kylrth:
Deployed to https://libreddit.kylrth.com, works as promised

mikupls:
Thanks for the work!

Does it work though?
I get `Bad response from reddit.com` when I try to use a share link.

My understanding is that `Url::parse` expects an absolute path, but we give it a relative one in the current code.
I had to apply this change to make it work for me: https://github.com/mikupls/libreddit/commit/9637a4afdc091255db2774edd7778765ddbaaf2a
This is not a nice solution because now the libreddit url is cluttered with share parameters (seemingly what `parsed_path.path()` tried to prevent).

pts314:
Thanks for the digging! I realized that we can just split after we have just the path, so that's what it's doing now

kylrth:
Just deployed this to https://libreddit.kylrth.com/, appears to work

skydiver1980 (approved):

skydiver1980 (approved):

ccuser44 (approved):

